PR #14, opened 2020-02-06
DAG Rehab

david-murrell:
This is the first of several DAG refactorings that need to occur. Much code cleanup is left, but there are several improvements I want to launch that are in this branch, notably:

1. Improved Vertex implementation that eliminates some "non-deterministic" edge traversals -- these were correct, but because edge maps were keyed by Edge*, the detected cycle could be different from run to run

2. EdgeFactory -- similar to VertexFactory in spirit

3. Added ability to dump Vertices and Edges to CSV files for import to some DAG exploration IDE's (such as Tulip)

4. Change misleading symbol and method names for GOP's

5. Improve the cohesion of DAG::sort()

6. Completely overhaul the DAG::CycleException outputters (writeText and writeDOT). The prior implementation was pretty convoluted.

7. Simplify a lot of the DAG cycle output; to make it easier for humans to read


sparta/sparta/kernel/EdgeFactory.hpp
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    void removeEdge(const Edge* e);
+    void dumpToCSV(std::ostream& os) const;
+
+private:
+    // I use bare Edge* here instead of std::unique_ptr<> since I need
+    // to be able to do a set::find() on the Edge address. The
+    // class destructor zaps the Edges
+    std::set<const Edge*> edges_;
```
klingaard: Shouldn't this be a set of `std::unique_ptr`?

sparta/sparta/kernel/Vertex.hpp
```diff
@@ -211,48 +222,54 @@ namespace sparta
...
-        const EMap & edges() const { return edges_; }
+        // const EdgeMap & edges() const { return outbound_edge_map_; }
+        const VertexList & edges() const { return outbound_edge_list_; }
         const Scheduleable * getScheduleable() const { return scheduleable_; }
         void setScheduleable(Scheduleable * s) { scheduleable_ = s; }
-        EMap::size_type numOutboundEdges() const { return edges_.size(); }
-        bool isOrphan() const { return ((num_inbound_edges_ == 0) && edges_.empty()); }
+        EdgeMap::size_type numOutboundEdges() const { return outbound_edge_list_.size(); }
+        bool isOrphan() const { return ((num_inbound_edges_ == 0) && outbound_edge_list_.empty()); }
         bool isInDAG() { return in_dag_; }
```
klingaard: This should be `const` method.

sparta/src/DAG.cpp
```diff
@@ -8,76 +8,98 @@
...
-            os << "\trankdir=TB;\n";
-            os << "\tnode [shape=record, fontname=Helvetica, fontsize=10];\n\n";
-        }
-        else {
-            os << "DAG CYCLE: " << std::endl;
+    /**
+     * Write a text version of the CycleException's cycle vertex list
+     * @param os
+     */
+    void DAG::CycleException::writeText(std::ostream& os) const {
```
klingaard: Nit: the method name `writeText` is a little vague.  How about something like `outputCycleIssueText`?

sparta/src/DAG.cpp
```diff
@@ -8,76 +8,98 @@
...
-                if(e != nullptr && dot) {
-                    os << " [fontsize=8 label=\"" << e->getLabel() << "\"]";
-                }
-                os << ";\n";
-                very_last = second;
+    /**
+     * Write a DOT graph version of the CycleException's cycle vertex list
+     * @param os
+     */
+    void DAG::CycleException::writeDOT(std::ostream& os) const {
```
klingaard: Same here -- deeper function name?  Also, can these methods be used to output the DAG generally?  I.e. not necessarily for a DAG issue?

sparta/src/Scheduleable.cpp
```diff
@@ -95,27 +123,33 @@ namespace sparta
...
         case SchedulingPhase::Tick:
-            dag->getGOPoint("Collection")->link(this->vertex_);
+            // dag->getGOPoint("Collection")->link(this->vertex_);
+            dag->link(dag->getGOPoint("Collection"), this->vertex_);
             this->precedes(dag->getGOPoint("Tick"));
             break;
         case SchedulingPhase::PostTick:
-            dag->getGOPoint("Tick")->link(this->vertex_);
+            // dag->getGOPoint("Tick")->link(this->vertex_);
+            dag->link(dag->getGOPoint("Tick"), this->vertex_);
```
klingaard: Can you remove the dead code?

sparta/src/Scheduler.cpp
```diff
@@ -108,7 +108,12 @@ Scheduler::Scheduler(const std::string& name, GlobalTreeNode* search_scope) :
     // start/stop times for differencing.
     sset_.setClock(scheduler_internal_clk_.get());
 
+    // TODO NOTE: Should we turn cycle-checking on by default (true)?
+    // Cycle-checking tells the DAG to flag a cycle at the earliest opportunity
+    // This means checking for a cycle as each edge is added to the DAG (DAG::link)
+    // It's expensive, but can save time in debugging precedence problems
```
klingaard: YES!  However, if you're concerned about the time spend, we can add it as a boolean to the `SimulationConfiguration` class and pass that value through to the Scheduler...

sparta/src/Vertex.cpp
```diff
@@ -10,133 +10,229 @@
...
+            outbound_edge_map_.erase(ei);
             sparta_assert(w->num_inbound_edges_ > 0);
             --(w->num_inbound_edges_);
+            efact.removeEdge(ei->second);
         }
         return true;
     }
 
-    void Vertex::assignConsumerGroupIDs(VList &zlist)
+#if 0
```
klingaard: Dead code?

sparta/src/VertexFactory.cpp
```diff
@@ -1,23 +1,28 @@
...
 
-#include "sparta/kernel/Vertex.hpp"
-#include "sparta/kernel/VertexFactory.hpp"
+#if 0
+VertexFactory::~VertexFactory() {
+    for (auto v : vertices_) {
+        delete(v);
+    }
+}
+#endif
```
klingaard: Safe to remove...

sparta/src/sparta.cpp
```diff
@@ -169,7 +169,7 @@ namespace sparta {
     constexpr char Scheduler::NODE_NAME[];
     const Scheduler::Tick Scheduler::INDEFINITE = 0xFFFFFFFFFFFFFFFFull;
     std::unique_ptr<sparta::SleeperThreadBase> SleeperThread::sleeper_thread_;
-    uint32_t Vertex::global_id_ = 0;
+    // uint32_t Vertex::global_id_ = 0;
```
klingaard: Dead code?

sparta/sparta/kernel/EdgeFactory.hpp
david-murrell: The thing that's preventing me from doing that (as was done for VertexFactory), is that I want to do a set::find(Edge*). If I were to use std::set(std::unique_ptr<const Edge*>) I apparently need to first convert the Edge* I'm looking for to a unique_ptr<> so that the two will compare. Sooooo, I just use bare pointers, and ~EdgeFactory() will do the deletions. I think there are probably other ways around this, but I chose to do this one out of simplicity.

sparta/sparta/kernel/Vertex.hpp
david-murrell: Yup, you're right

sparta/src/DAG.cpp
david-murrell: ok

sparta/src/DAG.cpp
david-murrell: No, these are just for the cycle. DAG has separate methods for full DAG output

sparta/src/Scheduleable.cpp
david-murrell: sure

sparta/src/Scheduler.cpp
david-murrell: that would be great!

sparta/src/Vertex.cpp
david-murrell: yes -- forgot to delete this

sparta/src/VertexFactory.cpp
david-murrell: yup, forgot about this one too

sparta/src/sparta.cpp
david-murrell: yup, i'll delete it

sparta/sparta/kernel/EdgeFactory.hpp
klingaard: I see.  One way to do this is to have two data structures -- `std::vector` of `std::unique_ptr` and the `set` that holds bare pointers.  But that's just a nit.

sparta/src/Scheduler.cpp
klingaard: Do you want me to do that in this branch or after the merge of the PR?  I think the latter...

sparta/sparta/kernel/EdgeFactory.hpp
david-murrell: Yes, I did think of that way around this too. I decided I was starting to go to too much trouble just to avoid a really simple destructor (and a statement in removeEdge). I do see the advantage to unique_ptr<>. but I couldn't find a way to leverage that which didn't require another data structure or passing around Edge id's to look up in a map

klingaard: ok

sparta/src/Scheduler.cpp
david-murrell: Yeah, if it's ok to wait for the PR merge, then you can just do in master or another tiny branch

david-murrell: (Guess we need another approver to sign off first, though)

PR #16, opened 2020-02-08
DataInPort precedence assertion

klingaard:
Added a DataInPort precedence assertion ensuring that a modeler does not put a precedence between two InPorts on difference phases.  Before this assert, the DAG would fire, providing a confusing message.

### DIFF
--- a/sparta/sparta/ports/Port.hpp
+++ b/sparta/sparta/ports/Port.hpp
@@ -409,6 +409,9 @@ namespace sparta
          */
         void precedes(InPort & consumer)
         {
+            sparta_assert(getScheduleable_().getSchedulingPhase() == consumer.getScheduleable_().getSchedulingPhase(),
+                          "ERROR: You cannot set precedence between two Ports on different phases: "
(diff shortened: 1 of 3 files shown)


sparta/sparta/ports/Port.hpp
```diff
@@ -409,6 +409,9 @@ namespace sparta
          */
         void precedes(InPort & consumer)
         {
+            sparta_assert(getScheduleable_().getSchedulingPhase() == consumer.getScheduleable_().getSchedulingPhase(),
```
avinabadasgupta: Shouldn't this sort of behavior from the modeler be a candidate for ```std::logic_error``` exception?

klingaard: 🤣 Yeah, we could argue that for most of the C++ exceptions in Sparta.  For now, I'm playing it safe ensuring the framework will catch this exception for the user and provide the standard error dump + backtrace...

avinabadasgupta: Fair enough.

klingaard:
Jack, Dave, Colby, can you look at this?

PR #20, opened 2020-02-14
Fixed issue with Queue entry allocation and deallocation

davidsifive:
The Queue allocates memory using 'nextPowerOfTwo_(num_entries*2)'. For example,
if 5 entries are requested, storage is allocated for 16 entries. The queue was
not managing the construction/destruction of the Queue entries properly allowing
all 16 entries to concurrently be valid, constructed objects even though the
user only request 5.  Fixed this by using a placement new when pushing an entry
and by explicitly calling the destructor when invalidating an entry.

### DIFF
--- a/sparta/sparta/resources/Queue.hpp
+++ b/sparta/sparta/resources/Queue.hpp
@@ -555,11 +555,10 @@ namespace sparta
          * any time for this data's position in the queue.
          * \warning appends through via this method are immediately valid.
          */
-        QueueIterator<false> push (const value_type & dat)
-        {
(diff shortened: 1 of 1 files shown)


klingaard:
Hey Avi, can you take a peek at this?

avinabadasgupta:
Looks good to me.

klingaard:
I think this is causing a memory leak in the CoreExample.  Investigating...

klingaard:
Yep.  If I revert this, the CoreExample runs to completion.  With this change, the CoreExample runs out of memory.  I think the in place new is bypassing the allocator somehow.

PR #21, opened 2020-02-21
Mike/simdb

MichaelSchoenfelder-SiV:
Sparta by default always emits a simdb data file.  The default type is sqlite, the default location is /tmp, and the name is auto-generated to be unique (good idea if /tmp is where we really wanted these things to go).  I put a one line change in Simultion.cpp to print out the location of this file to stdout so a user can figure out which one it is.

### DIFF
--- a/sparta/src/Simulation.cpp
+++ b/sparta/src/Simulation.cpp
@@ -1110,7 +1110,7 @@ void Simulation::saveReports()
             formatters_in_use = report_repository_->getFormattersByFilename();
             db::SimInfoSerializer serializer(SimulationInfo::getInstance());
             serializer.serialize(*stats_db_);
-	    std::cout << "SimDB file is: " << stats_db_->getDatabaseFile() << std::endl;
+            std::cout << "SimDB file is: " << stats_db_->getDatabaseFile() << std::endl;
         }
 
         auto saved_reports = report_repository_->saveReports();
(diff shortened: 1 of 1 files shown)


sparta/src/Simulation.cpp
```diff
@@ -1110,6 +1110,7 @@ void Simulation::saveReports()
             formatters_in_use = report_repository_->getFormattersByFilename();
             db::SimInfoSerializer serializer(SimulationInfo::getInstance());
             serializer.serialize(*stats_db_);
+            std::cout << "SimDB file is: " << stats_db_->getDatabaseFile() << std::endl;
```
klingaard: Does this print the location as well, or just the name?

MichaelSchoenfelder-SiV: The full path and file name.
```
Saving reports...
Automatic Simulation Summary [0,200001]  
  scheduler
          ticks                                                    = 200000 
          picoseconds                                       = 200000 
...
  [simdb] Validation of database-generated auto summary is currently unavailable
SimDB file is: /tmp/b487ac21-e0e9-4035-afcb-7a8618566367.db 
```
That line about "Validation of database..." is extraneous and probably should be deleted.

And, as I look at this more closely, I see I caused a side effect (print to stdout) in a lambda function definition.  I can move that statement to a later point where the lambda is used.  Can I withdraw my pull request?

jackjchen: You could git rebase -i to clean up commits, or you could switch the "base branch" of the PR, supposedly using the edit button next to the PR title.

klingaard: I can also close the pull request without merge.  Let me know what you want to do.

MichaelSchoenfelder-SiV: Never mind.  My brain took the afternoon off.  What I thought was happening isn't how lambdas work.  I think this code is OK.

PR #24, opened 2020-03-03
Fix pointer to member operator in Queue, Array

avinabadasgupta:


klingaard:
Awesome!  Can you fix the other resources too???

avinabadasgupta:
> Awesome! Can you fix the other resources too???

From what I say, Array is the only one left broken. Somehow, CircularBuffer and Buffer was already fixed. Not sure who did it. Working on Array now.

PR #28, opened 2020-03-04
Turned simdb off; reverted Queue allocation/deallocation fix

klingaard:
I normally don't like having two issues into one, but I needed to undo the Queue fix to get this change to regress.

### DIFF
--- a/sparta/src/CommandLineSimulator.cpp
+++ b/sparta/src/CommandLineSimulator.cpp
@@ -1953,7 +1953,7 @@ void CommandLineSimulator::populateSimulation_(Simulation* sim)
     //while so downstream SPARTA clients can revert to legacy reporting
     //infrastructure if they really need to.
     if (!feature_config_.isFeatureValueSet("simdb")) {
-        feature_config_.setFeatureValue("simdb", 1);
+        feature_config_.setFeatureValue("simdb", 0);
     }
     sim->setFeatureConfig(&feature_config_);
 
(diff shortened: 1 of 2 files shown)


sparta/src/CommandLineSimulator.cpp
```diff
@@ -1953,7 +1953,7 @@ void CommandLineSimulator::populateSimulation_(Simulation* sim)
     //while so downstream SPARTA clients can revert to legacy reporting
     //infrastructure if they really need to.
     if (!feature_config_.isFeatureValueSet("simdb")) {
-        feature_config_.setFeatureValue("simdb", 1);
+        feature_config_.setFeatureValue("simdb", 0);
```
cnyce: Why is this being disabled?  If it is not intended to be enabled again, I will split off SimDB into its own github asap.

klingaard: It's not being disabled, just turned off until it's needed.  We have a lot of testers that are running that are filling up /tmp with unused .db's.  We want to explicitly enable it on our full pipeline runs.

klingaard: You ok with this change?  There are two major reasons for wanting this change:
1. Frameworks shouldn't do quiet things in the background unless explicitly requested
2. We have numerous testers and we don't want to change all of them adding `--feature simdb 0`

cnyce:
I'm okay with disabling the database by default.

PR #31, opened 2020-03-06
Add rvalue version of push API

avinabadasgupta:


klingaard:
You know I'm gonna ask this -- can you add this to the other resources as well?

klingaard:
I tested this with software we're building at SiFive and it works!

avinabadasgupta:
> I tested this with software we're building at SiFive and it works!

avinabadasgupta:
> I tested this with software we're building at SiFive and it works!

Great. I will do the same for other resources.

Issue #32, opened 2020-03-07
Port_test is segfaulting on some builds


klingaard:
Issue in Port tests, specifically the fix to issue #15

klingaard:
Fixed.

PR #33, opened 2020-03-07
Fixed the Port tester; temporary deleted before DAG finalized

klingaard:


timsnyder:
Knute, you should mention that this Fixes #32 when you open your PR so that it will link these two together and make it easier to get to the change that fixed the issue.

klingaard:
No problem.  Good idea.

PR #34, opened 2020-03-07
Add continuous integration testing

timsnyder:
Uses conda environments on Azure pipelines, reusing [conda-forge infrastructure](https://conda-forge.org/docs/user/ci-skeleton.html).  Using conda because we need to support it for the users at the old place.

Results of CI are viewable at https://dev.azure.com/sparcians/map/_build?definitionId=1&_a=summary.  I've added a badge to the Readme.md that will show latest build results on master.

There are still several things remaining to be done but I'd like to get this initial PR merged:
- configure cmake/ctest to generate Test.xml output from `ctest --test-action test` so that it can be loaded into the Azure Pipelines test results.  This will provide per-test pass/fail trends across runs as well as better output for understanding failures.
- add valgrind tests.  There is an issue with the conda valgrind builds that needs to be sorted out and I didn't want to delay this any longer.
- upload final packages to anaconda.org
- build, test, package with DEBUG
- also include PYTHON stuff?  This will require reworking the hardcoded python37 version in places.
- build and deploy docs
- ensure that CI results are reported back into PRs
- ensure that CI is being run when we want it.  I think it may run for all pushes after this is merged.  
- Hook the results into the "Actions" tab of the github repo.  I know it's possible but I don't know if I've configured that yet.


timsnyder:
Looks like the CI builds are showing up in 'checks' on the PR.  That's good.

timsnyder:
I was hoping that rebasing my branch onto the fixes from Knute in #33 would resolve the failures I was seeing.  Still have some more.  I'm looking into them.

timsnyder:
The DAG_test error on linux is a segfault that looks like this:
```
Program received signal SIGSEGV, Segmentation fault.
sparta::Vertex::unlink (this=0xa1b1f0, efact=..., w=0xa1a820) at /home/conda/feedstock_root/build_artifacts/map_1583622171682/work/sparta/src/Vertex.cpp:50
50	                if (*el == ei->first) {
(gdb) where
#0  sparta::Vertex::unlink (this=0xa1b1f0, efact=..., w=0xa1a820) at /home/conda/feedstock_root/build_artifacts/map_1583622171682/work/sparta/src/Vertex.cpp:50
#1  0x000000000042d2ad in main () at /home/conda/feedstock_root/build_artifacts/map_1583622171682/work/sparta/test/DAG/DAG_main.cpp:144
```

Digging more into `el` and `ei`, gdb just says they are 'optimized out'.  Going to see if I can repo with a debug build.

timsnyder:
I'm not able to reproduce the `Report_triggers_test` failure when I run locally.  I will make the CI run do:
```
make regress -j $NUM_CPU || ctest -LE valgrind_test --output-on-failure
```
to at least see if the output of the test gives any useful information.

timsnyder:
The debug build repro'd DAG_test segfault and tells me there's something funky in the `el` iterator during `sparta::Vertex::unlink`:
```
Program received signal SIGSEGV, Segmentation fault.
0x0000000000510053 in sparta::Vertex::unlink (this=0x1d821f0, efact=..., w=0x1d81820)
    at /home/conda/feedstock_root/build_artifacts/map_1583622171682/work/sparta/src/Vertex.cpp:50
50	                if (*el == ei->first) {
(gdb) p el
$1 = {_M_node = 0x10ec8348e5894855}
(gdb) p ei
$2 = {_M_node = 0x1d82810}
(gdb) p ei->first
$3 = (sparta::Vertex * const) 0x1d81820
(gdb) p *el
$4 = (sparta::Vertex *&) <error reading variable>
(gdb) list
45	        if (ei == outbound_edge_map_.end()) {
46	            // Edge not present -- just ignore
47	            return false;
48	        } else {
49	            for (auto el = outbound_edge_list_.begin(); el != outbound_edge_list_.end(); ++el) {
50	                if (*el == ei->first) {
51	                    outbound_edge_list_.erase(el);
52	                }
53	            }
54	            outbound_edge_map_.erase(ei);
(gdb) p (*el == ei->first)
Cannot access memory at address 0x10ec8348e5894865
```

.ci_support/linux_.yaml
```diff
@@ -0,0 +1,25 @@
+boost_cpp:
+- 1.70.0
```
klingaard: Curious, is there a way to generate this file from the cmake configs?  I worry that this file and the required minimums on the build would get out of sync...

recipe/patches/0001-Fix-build-to-work-with-Anaconda-compiler-toolchain.patch
```diff
@@ -0,0 +1,39 @@
+From 6261d227b0cce2b765b3ca5e6dd661d6169c8f1e Mon Sep 17 00:00:00 2001
+From: Tim Snyder <snyder.tim@gmail.com>
+Date: Thu, 5 Mar 2020 17:17:17 -0600
+Subject: [PATCH] Fix build to work with Anaconda compiler toolchain
```
klingaard: Did you want this to be included in the patch?

timsnyder: Meh.  I could take it out but I just used the output of `git format-patch` without modifying them.  Is that stuff hurting anything?

I'd eventually like to have this incorporated into the cmake files but used a patch for now since it is only needed when using the conda toolchains.

.ci_support/linux_.yaml
timsnyder: This file is generated by `conda smithy rerender` run in the root of the repo.  It applies the boost version that is currently being used across all conda-forge recipes in https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/master/recipe/conda_build_config.yaml as well as any local overrides we have in `recipe/conda_build_config.yaml`.   If it does get out of sync, it will be caught in CI.

I agree that we should figure out one place to store the required minimums.  I think that it will be easier to have them stored in a yaml configuration file that cmake can look at rather than having to get other things to understand where the minimums are kept in our cmake files.

timsnyder:
Ok.  I have the DAG_test passing locally in linux.

Not being able to repro the macos failure is a real PITA.  I'm guessing that this might be caused by using a macos APK that will compile but has issues running some of our code.  I've just pushed changes that will try to build sparta with the APK from 10.13, 10.14 and 10.15.

timsnyder:
Awesome!  I fix one thing to have more stuff fail.  @klingaard do the [latest errors on linux](https://dev.azure.com/sparcians/map/_build/results?buildId=16&view=logs&j=1bf226d3-0e2f-52d8-fa93-7d9e633347b3&t=ec8d466d-e3b4-5115-4f06-222398f91e6c&l=883) look familiar to you at all?

klingaard:
> Awesome! I fix one thing to have more stuff fail. @klingaard do the [latest errors on linux](https://dev.azure.com/sparcians/map/_build/results?buildId=16&view=logs&j=1bf226d3-0e2f-52d8-fa93-7d9e633347b3&t=ec8d466d-e3b4-5115-4f06-222398f91e6c&l=883) look familiar to you at all?

Nope.  There're two errors there, and neither look familiar.  One looks like it's trying to print a `std::unique_ptr` and another complaining about a 0-arg collector getting at least one argument.  
What version of gcc is this?  (I'm being lazy and not looking for it in all of the output).

timsnyder:
> What version of gcc is this?
clang 9.0.1 backed by gcc 7.3

I'm using clang on linux because Benoy mentioned that they are only building with clang but then again, we're supporting the old place and they are using gcc.  I can make it build on linux with both of them.

klingaard:
> > What version of gcc is this?
> > clang 9.0.1 backed by gcc 7.3
> 
> I'm using clang on linux because Benoy mentioned that they are only building with clang but then again, we're supporting the old place and they are using gcc. I can make it build on linux with both of them.

I think I know what's going on.  This is related to @avinabadasgupta's change to support move operations within Queue.  Looks like gcc is instantiating all class types regardless of the fact they aren't being used.

avinabadasgupta:
> > > What version of gcc is this?
> > > clang 9.0.1 backed by gcc 7.3
> > 
> > 
> > I'm using clang on linux because Benoy mentioned that they are only building with clang but then again, we're supporting the old place and they are using gcc. I can make it build on linux with both of them.
> 
> I think I know what's going on. This is related to @avinabadasgupta 's change to support move operations within Queue. Looks like gcc is instantiating all class types regardless of the fact they aren't being used.

#38 Please review

recipe/patches/0001-Fix-build-to-work-with-Anaconda-compiler-toolchain.patch
jackjchen: Thanks @timsnyder incorporating the patch into the cmake files will help our local devs on RHEL7 and any on-premise CI flows too.

timsnyder:
closed and reopening to rerun the CI

jackjchen:
@timsnyder my local osx 10.15 build doesn't have the "57 - Report_triggers_test (Failed)", it is using boost 1.72 though in my local env.

timsnyder:
> @timsnyder my local osx 10.15 build doesn't have the "57 - Report_triggers_test (Failed)", it is using boost 1.72 though in my local env.

Thanks @jackjchen.  I had noticed that.  Turns out that the Boost cmake files on osx were also causing the static libs to be used.  I was able to work around that in the cmake so that we can build with 1.70 (which is what conda-forge is using for everything else and will be good to match).

timsnyder:
Ooh nice.  Somehow GH has gotten it's crap together and now sees the CI running again.

@klingaard I think I have this finally ready to merge 🤞 .  I want to wait for the latest CI to finish.  Hopefully it will be collecting test results for both platforms in [the test failure report UI](https://dev.azure.com/sparcians/map/_test/analytics?definitionId=1&contextType=build).  Either way, I want to merge it so that I have CI running and can utilize it for the Argos PR I want to open soon.  

The only possible downside is that failing CI might block merging other pull requests.

Issue #45, opened 2020-03-11
Conditionally compile Collectable::collect on ostream operator availability 


avinabadasgupta:
Most of the resource classes have internal iterable collectors which enable them for pipeline collection.
Currently, there is a strict restriction that the template type of resources must know how to respond to ostream operators.
This is cumbersome for cases where we need resources but do not care about pipeline collection.
We should not compile Collectable code that needs ostream operators if ostream operators are not defined by the template type.
This can be easily achieved with a mixture of constexpr if statement with has_ostream_operator metafunction in Utils.h.

klingaard:
@cnyce will be addressing this with a new SimDB.

PR #49, opened 2020-03-11
Conditionally compile Collectable::collect on ostream availability 

avinabadasgupta:
#45

### DIFF
--- a/sparta/sparta/collection/Collectable.hpp
+++ b/sparta/sparta/collection/Collectable.hpp
@@ -114,10 +114,12 @@ namespace sparta{
                 collected_object_ = collected_object;
 
                 // Get an initial value, if available
-                if(collected_object) {
-                    std::ostringstream ss;
-                    ss << *collected_object;
-                    prev_annot_ = ss.str();
+                if constexpr (sparta::utils::has_ostream_operator<std::decay_t<DataT>>::value) {
+                    if(collected_object) {
+                        std::ostringstream ss;
+                        ss << *collected_object;
+                        prev_annot_ = ss.str();
+                    }
                 }
             }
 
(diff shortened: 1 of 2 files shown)


klingaard:
Has `sparta::utils::has_ostream_operator` always been available?

avinabadasgupta:
> Has `sparta::utils::has_ostream_operator` always been available?

Not always. I added it to Utils when I was working on the ```EnumCycleHistogram``` as a part of the ```StateResidencyTracker``` development.

sparta/sparta/collection/Collectable.hpp
```diff
@@ -114,10 +114,12 @@ namespace sparta{
...
-                if(collected_object) {
-                    std::ostringstream ss;
-                    ss << *collected_object;
-                    prev_annot_ = ss.str();
+                if constexpr (sparta::utils::has_ostream_operator<std::decay_t<DataT>>::value) {
+                    if(collected_object) {
+                        std::ostringstream ss;
+                        ss << *collected_object;
+                        prev_annot_ = ss.str();
+                    }
```
klingaard: Wonder if there should be an `else` clause for determining if there is a `PairDefinition`.  Essentially, if this class is being used by a developer, they want _some kind_ of collection.   I think with the implementation here, it will quietly do nothing if neither is provided -- or am I missing something?

klingaard: If I am correct on my understanding, we should static assert that this class isn't usable the modeler -- unless they provide *some* means of collection.

klingaard:
Hey Avi, are you using `avi_branch_46` to do this work?  It was closed and squashed in a previous PR...

avinabadasgupta:
> Hey Avi, are you using `avi_branch_46` to do this work? It was closed and squashed in a previous PR... Shouldn't it be issue_49?

It should be avi_issue_45. I mistyped the branch name.

sparta/sparta/collection/Collectable.hpp
avinabadasgupta: > If I am correct on my understanding, we should static assert that this class isn't usable the modeler -- unless they provide _some_ means of collection.

This is targeted for use cases where we have a resource but we are not interested in pipeline collection. So, we should not really be bothered with providing ostream operators or anything else that is needed for a functioning collector. In such a situation, if we static assert, then we stop compiling. So we haven't gained anything. Yes, this will ignore the collection if there are no streaming methods, but that is on the developer. If they want pipeline collection, then they must define ostreams or go the NVP way. The whole purpose of this feature is to not make a modeler write extra code when they needed arrays, queues etc but do not need collection. In such a case, I feel we should not stop compiling.Resources and pipeline collector are two different components and I don't like this strict coupling between them. But, if you have concerns, I can remove this and make the modelers define ostreams. I just wanted to see if something like this can be done.

klingaard: Ah, this goes back to the question -- should the resources _expect_ to be able to perform pipeline collection.  The answer is no, they shouldn't -- and forcing this requirement on the modeler has always been a pain.

How about this -- let's go back to your idea of removing the `enableCollection` methods and forcing the modeler to use an external `IterableCollector` on them (yes, I'm going back on my original opinion).  Then, in this class, you can implement the if-elif-else assert.  What do you think?

klingaard: I rather the `Collectable` class not quietly ignore collection if the caller doesn't support `operator<<` nor `PairDefintion`

avinabadasgupta: I like this. I will remove all this stuff from resources and re-write certain parts in ```example_core``` and see if everything works.

avinabadasgupta: > I rather the `Collectable` class not quietly ignore collection if the caller doesn't support `operator<<` nor `PairDefintion`

After I remove iterables from resources, I will put the ```static_assert``` to assert that ostream operators are visible, because now, you really need that.

klingaard: Not sure I follow.  Are you saying that you will keep the `enableCollection` method and add a static assert?

avinabadasgupta: Yes. If we take out the internal iterable collector out of resources and make the modelers go the external way of creating iterable collectors, then the modelers must know how to enable collection. So, we can keep the ```enableCollection``` method inside resources and make the method accept an iterable collector as argument and reuse the same code which we have right now.

klingaard: Ah.  I was thinking about removing the method entirely, and just letting C++ "nature" take it's course.  Code would go from this:
```
   my_queue_.enableCollection(my_node);
```
to this:
```
    my_iterable_collector_(node, "MyQueueCollector", "My description", my_queue_, my_queue_.size);
```
where `my_iterable_collector_` is
```
    sparta::collection::IterableCollector<sparta::Queue> my_iterable_collector_
```

avinabadasgupta: Lot cleaner this way. I'll use this method.

klingaard: We could create specializations for `sparta::Queue, Array`, etc, etc of the `IterableCollector`

avinabadasgupta: I will take a stab at this.

Issue #53, opened 2020-03-15
SpartaSharedPointer doesn't support move


klingaard:
Will look at adding this support.  This need was sparked by the addition of move/forwarding semantics added to the Sparta resources.

Issue #48, opened 2020-03-11
Add watermark callback to SpartaSharedPointerAllocator


knute-sifive:
To the allocator class, add the ability to add a `SpartaHandler` or `std::function` that allows the modeler to be notified when the watermark on their allocator was hit.  This allows the developer to pause simulation and see what objects are outstanding for debug,

Issue #17, opened 2020-02-10
Sparta's memory allocator allows modeler's to shoot foot


knute-sifive:
Currently the `SpartaSharedPointerAllocator` will complain if it's being deleted and there are outstanding `SpartaShared` pointers hanging out in simulation.  This is good, but bad. Two things happen:

1. The outstanding memory objects are now orphaned.
1. The outstanding memory objects memory is deleted (i.e. they're gutted!)

Need to figure out a better mechanism to handle this problem.

klingaard:
After thinking about this for a bit, I had a "der" moment.  Pool allocators are static for a reason -- this Issue's reason.  Technically the pool allocator should be the last thing destructed in a program -- this guarantees all memory has either been returned (in a well-behaved program) or at least will no longer be accessed by a bad one.  This allocator is no different.  I will update the comments on expected and suggested uses for the allocator.

PR #54, opened 2020-03-15
SpartaSharedPointer: Fix for issue #17, #48, and #53

klingaard:
Fix for Issue #17: updated the comments to illustrate how to create a safe Allocator
Fix for issue #48: Added a means for a developer to add a watermark callback to the Allocator
Fix for Issue #53: Added move support for construction and assignment.

### DIFF
--- a/sparta/sparta/utils/SpartaSharedPointer.hpp
+++ b/sparta/sparta/utils/SpartaSharedPointer.hpp
@@ -97,7 +97,7 @@ namespace sparta
          * Construct a new Reference Pointer and take initial
          * ownership of the given object pointer.
          */
-        explicit SpartaSharedPointer(PointerT * p = nullptr) :
+        explicit SpartaSharedPointer(PointerT * p = nullptr) noexcept :
             ref_count_(new RefCount(p))
         { }
 
@@ -105,7 +105,7 @@ namespace sparta
          * \brief Constructor for SpartaSharedPointer<T> ptr = nullptr;
          * \param nullptr_t
          */
-        constexpr SpartaSharedPointer(std::nullptr_t) :
+        constexpr SpartaSharedPointer(std::nullptr_t) noexcept :
             ref_count_(new RefCount(nullptr))
         { }
 
@@ -115,7 +115,7 @@ namespace sparta
          *
          * The two reference pointers now share the common memory
          */
-        SpartaSharedPointer(const SpartaSharedPointer & orig) :
+        SpartaSharedPointer(const SpartaSharedPointer & orig) noexcept :
             memory_block_(orig.memory_block_),
             ref_count_(orig.ref_count_)
         {
@@ -129,7 +129,7 @@ namespace sparta
          * This SpartaSharedPointer now replaces orig.  Orig becomes a
          * nullptr with no references
          */
-        SpartaSharedPointer(SpartaSharedPointer && orig) :
+        SpartaSharedPointer(SpartaSharedPointer && orig) noexcept :
             memory_block_(orig.memory_block_),
             ref_count_(orig.ref_count_)
         {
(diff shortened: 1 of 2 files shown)


sparta/sparta/utils/SpartaSharedPointer.hpp
```diff
@@ -525,11 +631,7 @@ namespace sparta
                 else {
                     if(SPARTA_EXPECT_FALSE(allocated_ > water_mark_)) {
                         if(!water_mark_warning_) {
```
avinabadasgupta: Do you have a sense of how probable it would be for modelers to use this watermark feature?
I was thinking if we could use ```SPARTA_EXPECT_TRUE``` for the if clause at line 633 if it is not likely.

sparta/sparta/utils/SpartaSharedPointer.hpp
```diff
@@ -122,6 +122,22 @@ namespace sparta
...
         }
 
+        /**
+         * \brief Take ownership of a reference pointer
+         * \param orig The original pointer
+         *
+         * This SpartaSharedPointer now replaces orig.  Orig becomes a
+         * nullptr with no references
+         */
+        SpartaSharedPointer(SpartaSharedPointer && orig) :
```
avinabadasgupta: I am a bit unsure about this. Typically, move constructors and move assigns should be ```nothrow``` methods. 
Can we just make the ```ref_count_``` of the moved from pointer point to ```nullptr``` instead of allocating a new ```RefCount```?
```
orig.ref_count_ = nullptr;
```

sparta/sparta/utils/SpartaSharedPointer.hpp
klingaard: It's looking more and more likely as we're using it.  I'll add the macro.  Good point.

sparta/sparta/utils/SpartaSharedPointer.hpp
klingaard: Originally thought that, but then I have to add an `if` condition on all uses of `ref_count_` to make sure it isn't nullptr.  So, instead, I make the original look like it was constructed with `nullptr_t`.

klingaard: I'll investigate the `nothrow`.  Good catch.

sparta/sparta/utils/SpartaSharedPointer.hpp
klingaard: Done.

sparta/sparta/utils/SpartaSharedPointer.hpp
klingaard: Done.  Although I couldn't add it to all operators as some do a check to ensure the user isn't doing something dumb (and that code throws an exception).

klingaard:
Ok, now I'm worried.  Adding the move operators _tanked_ the performance of SpartaSharedPointer.  Why?!?

klingaard:
It went from taking 1.3 seconds to taking over 7 in the tester!

klingaard:
> Ok, now I'm worried. Adding the move operators _tanked_ the performance of SpartaSharedPointer. Why?!?
> 
> It went from taking 1.3 seconds to taking over 7 in the tester!

Ugh... I know why.  In the move operation, I'm allocating a new ref_count_.  You pointed this out, Avi, that this could be bad.  And... it is.  Will have to rethink the move operations.

avinabadasgupta:
> > Ok, now I'm worried. Adding the move operators _tanked_ the performance of SpartaSharedPointer. Why?!?
> > It went from taking 1.3 seconds to taking over 7 in the tester!
> 
> Ugh... I know why. In the move operation, I'm allocating a new ref_count_. You pointed this out, Avi, that this could be bad. And... it is. Will have to rethink the move operations.

Oh wow, that allocation completely broke your move. C++ is unforgiving.

PR #61, opened 2020-03-16
SpartaSharedPointer: completely nullified the moved SpartaSP object

klingaard:
Added asserts on accessors to enure a user is NOT working with a dead
SpartaSharedPointer.  This brought the speed of SpartaSharedPointer back
to fast.

### DIFF
--- a/sparta/sparta/utils/SpartaSharedPointer.hpp
+++ b/sparta/sparta/utils/SpartaSharedPointer.hpp
@@ -76,18 +76,18 @@ namespace sparta
         /// Unlink the reference and delete the memory if last to point to it
         void unlink_()
         {
-            if(SPARTA_EXPECT_FALSE(ref_count_ != nullptr))
+            // DO not put a check for ref_count_ != nullptr here (or
+            // an assert).  This function is called by the destructor,
+            // which already checks for this.
+            --ref_count_->count;
+            if(SPARTA_EXPECT_FALSE(ref_count_->count <= 0))
             {
-                --ref_count_->count;
-                if(SPARTA_EXPECT_FALSE(ref_count_->count <= 0))
-                {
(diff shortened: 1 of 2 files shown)


sparta/sparta/utils/SpartaSharedPointer.hpp
```diff
@@ -68,20 +68,26 @@ namespace sparta
...
-            if(--ref_count_->count == 0) {
-                if(memory_block_) {
-                    memory_block_->alloc->release_(memory_block_);
-                }
-                else {
-                    delete ref_count_;
-                    ref_count_ = 0;
+        void unlink_()
+        {
+            if(SPARTA_EXPECT_FALSE(ref_count_ != nullptr))
```
avinabadasgupta: Why don't you assert here as well instead of the branching?

klingaard: I suppose I could move this check to the destructor of SpartaSharedPointer.  That's where this `if` statement means something.  I'll try that.

klingaard: Also, this function is called by the destructor, which shouldn't throw

avinabadasgupta: Okay, then you definitely need the check.

klingaard: Actually, I don't.  I can move it safely to the destructor.  All of the functions in SpartaSharedPointer perform this check (in an assert) before calling unlink.  So, good catch.  ;)

avinabadasgupta: Nice, this might add a little bit of performance in release mode.

PR #63, opened 2020-03-16
Found and fixed a bug in SpartaSharedP's refcounter

klingaard:
The count is supposed to be 0 if the pointer is nullptr, but this is a pain
to think about copies, moves, and assignments.  Put it back to using a positive count
and simply have use_count return 0 for nullptr pointer assignments.  Turns out, this is
what shared_ptr does.


klingaard:
Sorry, Avi, the team is blocked on this, gonna force a rebase and merge, but will still value your input.

avinabadasgupta:
> Sorry, Avi, the team is blocked on this, gonna force a rebase and merge, but will still value your input.

Looked good to me.

Issue #60, opened 2020-03-16
Queue, Buffer, and Pipeline should use in-place new


davidsifive:
I think there is a bug in sparta Queue, Buffer, and Pipeline.  These structures should use in-place new when allocating a new entry and call the destructor of an entry when it goes out of scope.  This would ensure that an 18 entry queue, for example, even if it allocates space for something like 64 entries in the implementation will only have 18 entries alive at any one time.  I tried to fix this in Queue (I believe it was Queue) but Knute said it was causing some issue.  Knute had to back out my changes from Sparta.  I never got back to this.  In my situation the 10 entry allocator for my 5 entry queue was aborting.

From Knute:
Yeah, I had to back this out — there were random seg faults in the Sparta testers using Queue. Destruction of popped, out or reach elements seems to be a better solution than the in-place new (possibly).

klingaard:
Hey Avi, David tried to make a fix to this in the past, but I ran into seg faults with the examples.  Didn't have time to track it down.  Anyway, this is the issue explained out:

For performance, the resources will allocate up to twice as much memory (up to the next pow2) for the resource.  When items _inside_ the resource are popped or erased or invalidated, they remain "alive" until they are replaced by the user.  David wants them killed immediately.  He's seeing the issue with his allocators complaining that he's "run out of memory" when using an Allocator coupled with a Buffer or Queue.  As an example, if he has a buffer with 5 entries, technically he only needs an allocator for roughly 6 objects -- 5 "alive" in the buffer and 1 to replace on of those objects.  But since the Buffer doubles the size internally or speed, he needs 11.

avinabadasgupta:
Will look into this.

klingaard:
Thanks.  I think all we need is invalidation of the object on erase/pop/etc.  I don't think we need the in place new, but could be wrong.

klingaard:
This was David's original commit: 0bd92035dd8ca44a364f55d6b6cf548274796b8b

davidsifive:
The Queue allocates memory for next_pow_2(2*num_entries).  So, when I ask for 5 it allocates 16.  My allocator needs to be greater than 16 to not die.

Thanks for looking into this Avi.
David

avinabadasgupta:
I think we should be allocating an array of unintialized elements in the Queue during its construction, in order to use placement new during push method. Currently, the queue constructor uses new operator to default-initialize the internal array of elements. We should use malloc to non-default-initialize this array, in order to use in-place new construction of elements in push method.

[...]

PR #64, opened 2020-03-18
Queue In-place construction

avinabadasgupta:


avinabadasgupta:
Not to be merged right now.
This is just to show you the changes.
There is one more unrelated issue that needs to be addressed.

klingaard:
It looks like the Queue_test is failing...

avinabadasgupta:
> It looks like the Queue_test is failing...

Yes, but it's for a different reason.
From what I found out, you cannot really move something when using placement new operator.
So, one check is failing in the Queue test which tests if a moved string size is empty.

klingaard:
Gotcha.  Suggestion:  in the Queue_test, create an Allocator and use it with a Queue.  Technically, you should be able to create an Allocator with size == queue_capacity + 1 and it should never run out of memory.

avinabadasgupta:
> Gotcha. Suggestion: in the Queue_test, create an Allocator and use it with a Queue. Technically, you should be able to create an Allocator with size == queue_capacity + 1 and it should never run out of memory.

Okay, so create a Queue templatized on SpartaSharedPointer of something and also use SpartaSharedPointerAllocator along with that, but make the max block size of this allocator to be queue_capacity + 1.

klingaard:
Yessir

avinabadasgupta:
Fix for #67 and Queue portion of #60.

sparta/sparta/resources/Queue.hpp
```diff
@@ -404,8 +398,14 @@ namespace sparta
...
             vector_size_(nextPowerOfTwo_(num_entries*2)),
+
             name_(name),
-            collector_(nullptr)
+            collector_(nullptr),
+
+            // Make the queue twice as large and a power of two to
+            // allow a complete invalidation followed by a complete
+            // population
+            queue_data_((value_type *)malloc(sizeof(value_type) * vector_size_))
```
klingaard: So what issues did you run into using `new`?

avinabadasgupta: We definitely don't want to use new operator here because we don't want initialized memory. We want uninitialized memory so that we can placement-new construct as we get data.

klingaard: Actually, I was curious why not use `new char[]` with the `unique_ptr` that was there before, but ensure the pointer is aligned to a `DataT` address.  I have no problem, really with the malloc, except for the case where the Queue constructor throws, which there will be a memory leak...

avinabadasgupta: I did think of using ```std::aligned_storage``` for this.
I will give it a shot.

avinabadasgupta: Went along a different route.
Used ```unique_ptr``` with ```malloc``` and custom deleter which does ```free```.

davidsifive: Does this mean that value_type no longer needs a default constructor?

avinabadasgupta: I believe so, yes.

avinabadasgupta: In my test, I just removed the default ctor for value_type and works as expected.

sparta/sparta/resources/Queue.hpp
```diff
@@ -626,12 +642,18 @@ namespace sparta
         const_iterator end() const { return const_iterator(this,false);}
 
     private:
+        struct FreeToDelete_{
```
klingaard: ... or is is DeleteToFree?  :)

klingaard:
Well, since you've started this... 😛 

Can I do this?
```
   class { 
   public:
       MyObj(int) {}
    };

   sparta::Queue<MyObj> my_queue;
   my_queue.push(10);
}
```

sparta/sparta/resources/Queue.hpp
avinabadasgupta: Good observation, changing it.

avinabadasgupta:
> Well, since you've started this... 😛
> 
> Can I do this?
> 
> ```
>    class { 
>    public:
>        MyObj(int) {}
>     };
> 
>    sparta::Queue<MyObj> my_queue;
>    my_queue.push(10);
> }
> ```

I believe yes, since you haven't made the constructor ```explicit```.

klingaard:
> > Well, since you've started this... 😛
> > Can I do this?
> > ```
> >    class { 
> >    public:
> >        MyObj(int) {}
> >     };
> > 
> >    sparta::Queue<MyObj> my_queue;
> >    my_queue.push(10);
> > }
> > ```
> 
> I believe yes, since you haven't made the constructor `explicit`.

Ahhhh!!  You got me there. 

How about THIS?
```
{
   class { 
   public:
       MyObj(int, int) {}
    };

   sparta::Queue<MyObj> my_queue;
   my_queue.push({10, 10});
}
```

avinabadasgupta:
> > > Well, since you've started this... 😛
> > > Can I do this?
> > > ```
> > >    class { 
> > >    public:
> > >        MyObj(int) {}
> > >     };
> > > 
> > >    sparta::Queue<MyObj> my_queue;
> > >    my_queue.push(10);
> > > }
> > > ```
> > 
> > 
> > I believe yes, since you haven't made the constructor `explicit`.
> 
> Ahhhh!! You got me there.
> 
> How about THIS?
> 
> ```
> {
>    class { 
>    public:
>        MyObj(int, int) {}
>     };
> 
>    sparta::Queue<MyObj> my_queue;
>    my_queue.push({10, 10});
> }
> ```

I do not think this would work.
The compiler would probably complain that it cannot convert from ```initializer_list``` to ```MyObj```.
However, if you remove the constructor entirely, it will work I believe because of aggregate initialization.

klingaard:
Meh.  Food for thought.

klingaard:
@davidsifive , can you checkout this branch and see if it still works with SiFive's simulator?

davidsifive:
OK, will do.

davidsifive:
Three of the regression tests fail.  They hit an assert in SpartaSharedPointer.hpp::reset().

Here is the method:

        void reset(PointerT * p = nullptr) {
            sparta_assert(ref_count_ != nullptr, "This is a dead SpartaSharedPointer");
            unlink_();
            ref_count_     = new RefCount(p);
            memory_block_  = nullptr;
        }

If I change it to:

        void reset(PointerT * p = nullptr) {
            if (ref_count_ != nullptr) {
                unlink_();
            }
            ref_count_     = new RefCount(p);
            memory_block_  = nullptr;
        }

regression passes.  It seems perfectly reasonable to reset a dead SpartaSharedPointer.

Should I commit the change?  How do I run sparta regression?

avinabadasgupta:
> Three of the regression tests fail. They hit an assert in SpartaSharedPointer.hpp::reset(). Here is the method: void reset(PointerT * p = nullptr) { sparta_assert(ref_count_ != nullptr, "This is a dead SpartaSharedPointer"); unlink_(); ref_count_ = new RefCount(p); memory_block_ = nullptr; } If I change it to: void reset(PointerT * p = nullptr) { if (ref_count_ != nullptr) { unlink_(); } ref_count_ = new RefCount(p); memory_block_ = nullptr; } regression passes. It seems perfectly reasonable to reset a dead SpartaSharedPointer. Should I commit the change? How do I run sparta regression?
> […](#)

```
mkdir debug/release; 
cd debug/release
cmake .. -DCMAKE_BUILD_TYPE=Debug/Release
make regress
```

klingaard:
I purposely decided to not allow a dead pointer to be reset.  But it looks like `shared_ptr` allows this, so it's a valid fix...

avinabadasgupta:
Should in-place new construction be the way to move forward for every other resource class?

davidsifive:
Sparta regression passes with this change.  Should I commit the code?  Do I have permission to commit code?  Should I push it on a branch?

avinabadasgupta:
> Sparta regression passes with this change. Should I commit the code? Do I have permission to commit code? Should I push it on a branch?

Can you push it on a branch and open a PR?

klingaard:
> Should in-place new construction be the way to move forward for every other resource class?

Yep.  I think so.

klingaard:
> > Sparta regression passes with this change. Should I commit the code? Do I have permission to commit code? Should I push it on a branch?
> 
> Can you push it on a branch and open a PR?

I'll do this.

klingaard:
> Sparta regression passes with this change. Should I commit the code? Do I have permission to commit code? Should I push it on a branch?

Don't worry about it.  I'll do it.

Issue #68, opened 2020-03-19
It'd be handy to allow use of `const` methods for CREATE_SPARTA_HANDLER


klingaard:
Now, I'd love to abandon the macro completely and just use some cool C++17 feature, but for now it'd be handy to be able to create a SpartaHandler with a const method:

```
class Foo
{
private:
    void myMethod_() const {}
};

Foo::Foo()
{
    auto func = CREATE_SPARTA_HANDLER(Foo, myMethod_);
}
```
You get this when you try this:
```
error: no matching function for call to 'from_member'
             CREATE_SPARTA_HANDLER(Foo, myMethod_)};
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 note: expanded from macro 'CREATE_SPARTA_HANDLER'
    sparta::SpartaHandler::from_member<clname, &clname::meth>            \
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 note: candidate template ignored: invalid explicitly-specified argument for template parameter 'TMethod'
        static SpartaHandler from_member(T* object_ptr, const char * name = "")
```

klingaard:
Background task for you, Avi.  This one would be kind of fun to tackle.

avinabadasgupta:
Which one?
Adding const methods to SpartaHandler or removing the macro completely?

avinabadasgupta:
To add the const methods, shouldn't it just be this variation
```
#define CREATE_SPARTA_HANDLER(clname, meth)                                    \
    sparta::SpartaHandler::from_member<clname, &clname::meth>            \
    (this, #clname"::"#meth"()")
```
to
```
#define CREATE_SPARTA_HANDLER_CONST(clname, meth)                       \
    sparta::SpartaHandler::from_member<clname, &clname::meth>             \
    (this, #clname"::"#meth"() const")
```

klingaard:
I dunno.  :)  Maybe...  Getting rid of the macros would be the next challenge!

Basically, change this:
```
a_delay_in_.registerConsumerHandler
    (CREATE_SPARTA_HANDLER_WITH_DATA(MyDevice, myDataReceiver_, uint32_t));
```
to this:
```
a_delay_in_.registerConsumerHandler(myDataReceiver_);
```

[...]

PR #71, opened 2020-03-20
CREATE_SPARTA_HANDLER const methods

avinabadasgupta:
#68

### DIFF
--- a/sparta/sparta/kernel/SpartaHandler.hpp
+++ b/sparta/sparta/kernel/SpartaHandler.hpp
@@ -66,7 +66,28 @@ namespace sparta
             return d;
         }
 
+        template <class T, void (T::*TMethod)() const>
+        static SpartaHandler from_member(T* object_ptr, const char * name = "")
+        {
+            SpartaHandler d(name);
+            d.object_ptr = object_ptr;
+            d.stub_ptr = &method_stub<T, TMethod>;
+            d.arg_count_ = 0;
+            return d;
+        }
+
         template <class T, void (T::*TMethod)(), void (T::*TMethodClear)()>
+        static SpartaHandler from_member_clear(T* object_ptr, const char * name = "")
+        {
+            SpartaHandler d(name);
+            d.object_ptr = object_ptr;
+            d.stub_ptr = &method_stub<T, TMethod>;
+            d.clear_ptr = &method_stub<T, TMethodClear>;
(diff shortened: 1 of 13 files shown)


sparta/sparta/kernel/SpartaHandler.hpp
```diff
@@ -251,6 +316,45 @@ namespace sparta
...
+     *     }
+     *
+     *     void myMethod() {
+     *         std::cout << "myMethod called" << std::endl;
+     *     }
+     *
+     * };
+     * \endcode
+     */
+#define CREATE_CONST_SPARTA_HANDLER(clname, meth)                            \
```
klingaard: Did you really need a separate macro?

avinabadasgupta: No. Got confused there for a bit. Changing.

avinabadasgupta: My latest commit is not really what you wanted but it's a start.
I removed the need for Sparta Handlers from registerConsumerHandler method of Ports.
Tested this on core example.

klingaard:
So, this is a pretty big change.  Not sure if it will fly with folks.  But, I'm gonna challenge you once more...  I'll bet you can figure out the argument type of the function as well as determine the address of the function handed.  So you can go from this:
```
in_mss_ack_sync_.registerConsumerHandler<BIU, bool, &BIU::getAckFromMSS_>(this);
```
to this:
```
in_mss_ack_sync_.registerConsumerHandler(this, getAckFromMSS_);
```
After all, weren't you able to figure this out with your `PairDefinitions`?  :)

avinabadasgupta:
> So, this is a pretty big change. Not sure if it will fly with folks. But, I'm gonna challenge you once more... I'll bet you can figure out the argument type of the function as well as determine the address of the function handed. So you can go from this:
> 
> ```
> in_mss_ack_sync_.registerConsumerHandler<BIU, bool, &BIU::getAckFromMSS_>(this);
> ```
> 
> to this:
> 
> ```
> in_mss_ack_sync_.registerConsumerHandler(this, &BIU::getAckFromMSS_);
> ```
> 
> After all, weren't you able to figure this out with your `PairDefinitions`? :)
> 
> BTW, the Port knows the expected type for the argument. There's a hint there.

Let me play with this for a little more.

PR #91, opened 2020-04-04
Fixed issues with Registers not allowing large bitfields.

klingaard:
Re-wrote the regsiter bit field mask generator to properly calculate the
register mask.  Added testing for the specific issue found.


klingaard:
All regressions passed and we need this fixed, so I'm forcing a rebase

PR #92, opened 2020-04-04
Removed the old GCC_VERSION flag on CircularBuffer

klingaard:
Added testing for CircularBuffer const_iterator as well.


klingaard:
All regressions passed and we need this fixed, so I'm forcing a rebase

PR #104, opened 2020-04-15
Initialize Register::Definition by value instead of by reference

kathlenehurt-sifive:
In the Register class's constructor, Register::Definition is initialized by reference and a copy of the register definition object is not made. This requires that the register definitions be maintained somewhere outside of the Register class. 

To be able to initialize a Register using a definition read in from a file, all of the variables used to initialize that register have to be held onto. This use case could be simplified if Register::Definition was initialized by value so that the Register class maintained a copy of the register definition data itself.

### DIFF
--- a/sparta/sparta/functional/Register.hpp
+++ b/sparta/sparta/functional/Register.hpp
@@ -704,7 +704,7 @@ public:
      * marked as read-only regardless of any read-write fields containing those
      * bits
      */
-    RegisterBase(TreeNode *parent, const Definition def)
+    RegisterBase(TreeNode *parent, const Definition & def)
         : TreeNode(nullptr,
                    NULL_TO_EMPTY_STR(def.name),
                    NULL_TO_EMPTY_STR(def.group),
@@ -751,7 +751,7 @@ public:
         }
 
         // Add all fields
-        for(const auto & fdp : def_.fields){
+        for(auto & fdp : def_.fields){
             addField(fdp);
         }
 
@@ -949,7 +949,7 @@ public:
      * \return References to the same const Definition object supplied at
      * construction.
      */
-    const Definition getDefinition() const {
+    const Definition& getDefinition() const {
         return def_;
     }
 
@@ -1353,7 +1353,7 @@ private:
     }
 
     /*!
-     * \brief Register definition given at construction (not copied)
+     * \brief Register definition given at construction
      */
     const Definition def_;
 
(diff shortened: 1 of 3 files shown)


sparta/sparta/functional/Register.hpp
```diff
@@ -704,16 +704,16 @@ class RegisterBase : public TreeNode
      * marked as read-only regardless of any read-write fields containing those
      * bits
      */
-    RegisterBase(TreeNode *parent, const Definition *def)
+    RegisterBase(TreeNode *parent, const Definition def)
```
kathlenehurt-sifive, line 707: @knute-sifive, in this constructor does it make more sense to reference `def` or `def_` in the body? I changed the Field and aliases initialization to use `def_` since they need a reference to it, but left the rest alone.

My question is really more regarding style/convention... looks strange to me to have both `def` and `def_` used in the body.

klingaard, line 707: To avoid the implicit copy, make the second parameter a `const &`

sparta/sparta/functional/Register.hpp
```diff
@@ -930,28 +929,28 @@ class RegisterBase : public TreeNode
...
 
     /*!
      * \brief Gets the definition supplied during the construciton of this
      * Register.
      * \return References to the same const Definition object supplied at
      * construction.
      */
-    const Definition& getDefinition() const {
-        return *def_;
+    const Definition getDefinition() const {
```
klingaard, line 952: Return as a `const &` to avoid the copy.

sparta/sparta/functional/Register.hpp
```diff
@@ -727,37 +727,36 @@ class RegisterBase : public TreeNode
...
+        if(!isPowerOf2(def.bytes) || def.bytes == 0){
             throw SpartaException("Register \"")
                 << getName() << "\" size in bytes must be a power of 2 larger than 0, is "
-                << def->bytes;
+                << def.bytes;
         }
 
         // Add all fields
-        for(auto & fdp : def->fields){
+        for(const auto & fdp : def_.fields){
```
kathlenehurt-sifive, line 754: Didn't mean to add `const` here...

sparta/sparta/functional/Register.hpp
```diff
@@ -1356,7 +1355,7 @@ class RegisterBase : public TreeNode
     /*!
      * \brief Register definition given at construction (not copied)
```
kathlenehurt-sifive, line 1356: Need to update this comment!

PR #107, opened 2020-04-18
Changed all include guards to #pragma once

klingaard:
The original guards were not compliant.  This change is a bit risky, but I think most compilers will be fine with it.


avinabadasgupta:
Massive. Thanks for doing this.

klingaard:
Yer welcome.  Took me 20 minutes with a fancy sed script.  ;)

Issue #100, opened 2020-04-09
SpartaSharedPointer on nullptr_t or default construction allocates refcount


klingaard:
For performance, we don't need to be doing that.   The constructor and the `constexpr` nullptr_t can simply leave the internal refcount null.  Need to try and test this.

PR #108, opened 2020-04-18
On default construction of SpartaSharedPointer, do not allocate a ref_count_

klingaard:
Fixes a performance issue in SpartaSharedPointer for `nullptr` `SpartaSharedPointer` types.

Regression is clean and SiFive tools regressed fine.


klingaard:
resolves #100

PR #124, opened 2020-05-02
Argos/Sparta/Python updates/fixes

klingaard:
closes #123 
closes #121 

* Adds the transaction search program
* Fixes python2 -> python3 issues
* Adds more packages to the conda environment
* Fixes layout issues in Argos
* Fixes a bug in the Argos Outputter in Sparta
* Adds the argos database dumping utility

PR #126, opened 2020-05-03
Knute/125 python pr issues

### LINKED ISSUE #125: Address issues in PR #124 + other python issues
Full sparta python support is busted -- missing developer libraries in the conda as well as py2->py3 issues in sparta proper.

Also, this issue will address the comments in PR #124

klingaard:
Addresses issues from previous PR #124 

Fixes #125 .


sparta/CMakeLists.txt
```diff
@@ -110,7 +110,7 @@ if (NOT rc EQUAL "0")
 endif ()
 
 if (USE_LTO)
-  set (LTO_FLAGS "-flto -Ofast")
+  add_compile_options (-flto -Ofast)
```
criusx, line 113: hmm, is this on for all variants? Release, Debug and RelWithDebInfo?

klingaard: Yeah, probably not.  I think I'll remote this for sparta.  The simulators that use it can define it on their own...

criusx: You can specialize based on [CMAKE_BUILD_TYPE](https://cmake.org/cmake/help/v3.0/variable/CMAKE_BUILD_TYPE.html), so debug gets -O0 and -ggdb3 and RelWithDebInfo has O3.

klingaard: Yep.  Did that on an internal project.  I'm hesitant to play around with it here.  In my experiments, LTO didn't buy much at the Sparta level, but definitely at the simulation level.  Mostly because Sparta is headers...

criusx: seems fair, I guess I was thinking whether sparta's release variant got O3 or O2, the cmake default. in either case I'm not worried : )

conda.recipe/meta.yaml
```diff
@@ -39,13 +39,14 @@ requirements:
...
     - rsync
-    # add runtime requirements for argos to make it easier to script 
+    # add runtime requirements for argos to make it easier to script
     # creation of a 'dev' environment that can also run Argos
     - matplotlib-base
     - numpy
     - pillow
     - pyyaml
     - wxpython
+    - boost
```
timsnyder, line 49: Because of the way my janky environment creation scripts work right now, we need to duplicate `boost` in the run requirements and `ipython` in the build requirements.

helios/pipeViewer/transactionsearch/src/transaction_search.cpp
```diff
@@ -18,7 +18,7 @@
 #define NUMBER_OF_PROGRESS_UPDATES 50
 
 /*! Stores regular expression for comparison in callbacks */
-boost::regex global_regular_expression;
+std::regex global_regular_expression;
```
timsnyder, line 21: 💯 for switching to `std::regex`

sparta/test/TestingMacros.cmake
```diff
@@ -61,3 +60,15 @@ macro (sparta_recursive_copy build_target cp_file)
     add_custom_command (TARGET ${build_target} PRE_BUILD
       COMMAND cp -r ${CMAKE_CURRENT_SOURCE_DIR}/${cp_file} ${CMAKE_CURRENT_BINARY_DIR}/)
 endmacro (sparta_recursive_copy)
+
+#
+# Create a sparta test executable
+#
+macro (sparta_add_executable target)
```
timsnyder, line 67: Is this just for tests?  If so, can you name it `sparta_add_test_executable`?  Or, I think it would be better to change the comment to say this is a helper for adding an executable that needs to link against `${Sparta_LIBS}`.  Could be anything, doesn't have to be a test.  If you change the comment, this macro should probably be defined wherever `${Sparta_LIBS}` is defined.

sparta/test/TestingMacros.cmake
```diff
@@ -61,3 +60,15 @@ macro (sparta_recursive_copy build_target cp_file)
...
+  set (srcs "")
+  foreach(src_file ${ARGN})
+    list (APPEND srcs ${src_file})
+  endforeach()
+  add_executable (${target} ${srcs})
```
timsnyder, line 72: Is this the same as:
```
add_executable(${target} ${ARGN})
```

Not that what you wrote is hard to understand, it's just a bit verbose if using `${ARGN}` like that will work.

conda.recipe/meta.yaml
klingaard: Done.  
1. Copied `ipython` to the `build` requirements
1. Copied `boost` to the `run` requirements

sparta/test/TestingMacros.cmake
klingaard: Done.  That's was stupid of me.

sparta/test/TestingMacros.cmake
klingaard: Done -- renamed the macro.  Will push the changes in a minute.

I thought about the "anything" concept, but I don't want to force people that use Sparta into a corner where the default args for `add_executable` are not sufficient for their use.

helios/pipeViewer/transactionsearch/src/transaction_search.cpp
klingaard: Joe's idea.

helios/pipeViewer/argos_dumper/CMakeLists.txt
```diff
@@ -1,11 +1,11 @@
...
 # Enable testing
 enable_testing ()
 
-add_executable(Argos_test ArgosCollection_test.cpp)
-add_test (NAME Argos_test_RUN COMMAND Argos_test)
+add_executable(Argos_dumper ArgosCollection_test.cpp)
+add_test (NAME Argos_dumper_RUN COMMAND Argos_dumper)
 
-target_link_libraries (Argos_test ${Sparta_LIBS})
+target_link_libraries (Argos_dumper ${Sparta_LIBS})
```
knute-sifive, line 9: I forgot to complete finish the naming.  I'll do that in a following PR

PR #131, opened 2020-05-07
Argos now properly works with dabble pipeline

TopChangwatchaiSiFive:
Primary changes:
- Fixed bugs so ArgosDumper tool now works properly
- Fixed display issues so instruction data is properly formatted
- Added display_ID field to transaction_t

Details:
-     Fixed argos_dumper's output of pair ops
-     Database reader fixes certain pair integer values
-     Renamed sout to os in Database_dumper.cpp for naming consistency
-     Cleaned up annotation string
-     Changes db Reader.hpp:
    - pairid string is now unique ID, not blank
    - in addition to pc, pred_target and vaddr are hex fields
    - get rid of terminating null for string values
    - add else clause to add dummy string
-     In transactiondb reader, fixed double push onto valueVector
-     Added display_ID field to transaction data structure.  In addition to updating many interface functions:
    - initialized this field to bad value 0x1000 (it should be at most three hex digits) - search for BAD_DISPLAY_ID
    - if the special field named "DID" is the first in a pair vector, it is used as the display ID.  See writeRecord_() in Collectable.hpp
    - if this field contains a good value, it is prepended to the display string instead of the transaction ID.  
- See getAnnotation() in transactiondb.pyx
    - size of transaction_t has increased from 40 to 48 bytes
-     Suppress display of the DID field
-     Fixed automatic display of simple numbers
    - numbers are now formatted with a "reason" display ID
    - fixed display of "reason" display ID

-     Added special formatting for instructions to put key fields at the beginning


helios/pipeViewer/transactiondb/src/transactiondb.pyx
```diff
@@ -241,24 +247,51 @@ cdef class Transaction(object):
...
+                    value_string = "u" + format(int(my_value) % 10000, '>4d') + " "
+                    py_str_preamble += value_string.encode('utf-8')
+
+                elif my_name == "pc":
+                    value_string = "0x" + format(int(my_value, 16) & 0xffff, '>04x') + " "
+                    py_str_preamble += value_string.encode('utf-8')
+
+                elif my_name == "mnemonic":
+                    value_string = format(my_value[0:7], '<7') + " "
+                    py_str_preamble += value_string.encode('utf-8')
```
klingaard, line 286: Hard-coding this for now in the cython is ok for now, we need to get this going, but really we should make python stubs in the argos tool to handle this.

TopChangwatchaiSiFive: I totally agree.

PR #137, opened 2020-05-12
Updated pipeviewer readme

TopChangwatchaiSiFive:
Fixed markdown and clarified a step

### DIFF
--- a/helios/pipeViewer/README.md
+++ b/helios/pipeViewer/README.md
@@ -2,18 +2,21 @@
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


davidsifive:
Much better!  But the formatting is messed up if you view it at github.

TopChangwatchaiSiFive:
> 
> 
> Much better! But the formatting is messed up if you view it at github.

Fixed in PR #138 

I didn't realize that there are multiple incompatible versions of markdown.  The version in this PR works on the "official" version but Github apparently uses a nonstandard version.

So I've learned to always double-check the rendering in Github first.

Issue #133, opened 2020-05-08
Update Argos transaction data structure


TopChangwatchaiSiFive:
1. See if we can reduce the types of Argos transaction data structures to only two:  dictionaries and simple scalars
2. The key-value pairs in a transaction should be implemented as a dictionary of objects instead of separate lists of keys, sizes, and two types of values.

TopChangwatchaiSiFive:
Data structure also needs to include a field indicating whether this field should be displayed in hex format (like addresses).

PR #144, opened 2020-05-22
Added fields to be formatted as hex in Argos display

TopChangwatchaiSiFive:
This will be fixed in the future (perhaps part of #133), but for now certain fields are hardcoded to be displayed as hex.  In this PR, va, stf_va, and pa are added as hex fields.


TopChangwatchaiSiFive:
> Probably need to find a way to allow the end user to do this.

I already know what to do.  Just need to get around to it...too many fires currently.

Issue #128, opened 2020-05-04
Hard-coded size in Argos/pipeviewer


klingaard:
1. All of these hardcoded font sizes probably need to be refactored out into a per-user preferences file that can be reset to defaults (in an enhancement ticket, not for this PR).
1. If 'Monospace' needed a fallback in `layout_gui.py`, it will need it here too. (see the first item of this list)

_Originally posted by @timsnyder in https://github.com/sparcians/map/pull/124_

klingaard:
The last time I used argos, I could set the font sizes.  I think it's safe to say this issue is good for closure.

PR #153, opened 2020-05-28
Mike/issue 148 argos

### LINKED ISSUE #148: PipeViewer Argos no longer works on Ubuntu under gcc
Issues:

1. logsearch setup.py uses clang-only flags
1. The letters in the pipeline UI don't display
1. The fonts are too large (there is another issue suggesting that font sizes should be a preference; I think we changed them to look good on a Mac)
1. The Argos setup.py assume the compiler is defined by environment variables (CC) et al and don't honor what cmake is built with, e.g. `CC=clang CXX=clang++ cmake .. -DCMAKE_BUILD_TYPE=Release`

MichaelSchoenfelder-SiV:
This PR resolves #148 , plus cleans up some UI compatibility between Mac and Ubuntu.
It partially addresses issue #128 ; @bdutro added automatic font scaling based on display DPI.

Known issues:
1. Argos uses gcc on Ubuntu if CC et al envrionment variables point to gcc; this happens regardless of how cmake is configured.
2. Main layout window can't be resized on Ubuntu using NX X-Windows sessions.
3. Maximize and Close window decorations fail to appear on Ubuntu displaying to MobaXterm XServer in the native Windows window manager mode; the workaround is to close windows using the taskbar icon menu


MichaelSchoenfelder-SiV:
@bdutro resolved known issue number 3 and known issue number 2 went away.

PR #159, opened 2020-06-05
Knutel/issue 158 empty db error

### LINKED ISSUE #158: Argos Reader should be a little smarter on empty databases
It doesn't crash, but it doesn't complain really well either.  It says the database is corrupted, but in fact, it's empty.  A better error message is required.

knute-sifive:
Fixes a couple of issues:
1. If a user collects nothing in their simulation, even when enabled, the database is empty.  I made a change to have at least one record per collectable exist, even if empty
1. Added some friendlier messages from the Reader when it encounters an empty database
1. Added a collection test to eventually grow.
1. Added port collection testing

closes #158


sparta/sparta/collection/Collectable.hpp
```diff
@@ -303,10 +303,11 @@ namespace sparta{
                 sparta_assert(pipeline_col_ != nullptr,
                             "Collectables can only added to PipelineCollectors... for now");
 
-                if(collect && !prev_annot_.empty()) {
+                if(collect) {
```
bdutro, line 306: Have you checked this in Argos? I just want to double check that this doesn't cause every collected location to have a gigantic record stretching from time = 0.

knute-sifive: How I can check that?

bdutro: Run a sim, dump a DB, open it up and see whether the output makes sense. If every location has a record at the very beginning then something is probably wrong.

knute-sifive: Actually, would it make more sense to create a single open/closed empty record?

bdutro: It looks like your changes to the reader are already enough to properly handle an empty DB, so I don't really see why the empty record is necessary.

knute-sifive: Well, if someone collects a pipeout for one component in simulation (which never changes its value or even starts with one), and runs their entire simulation (1000s of cycles), the database will be empty instead of just blank for that component.  Not sure what the right behavior is.

bdutro: Hm, that's a good point. I think in that case the index file would still have heartbeats, so we could probably use that to distinguish a corrupt DB from an empty one.

klingaard: I don't feel comfortable with this change.  There isn't enough testing to see if I broke something downstream, and I barely remember why I originally wrote the code this way.  I'm going to revert the change to the Collectable, but keep the rest.

PR #163, opened 2020-06-23
Initialize the whole BitArray by data value

2henwei:
If `array_size` is greater than `data_size`, only the first `data_size` bytes will be initialized. In a zero-initialized scenario, the reset of BitArray are not correctly initialized. It would be nice to have it and doesn't influence the original functionality.

### DIFF
--- a/sparta/sparta/utils/BitArray.hpp
+++ b/sparta/sparta/utils/BitArray.hpp
@@ -64,7 +64,11 @@ namespace sparta {
                 }
 
                 initializeStorage_(roundUpTo_(array_size, sizeof(data_type)));
-                std::memcpy(data_, data, std::min(data_size, array_size));
+                size_t offset = 0;
+                do {
+                    std::memcpy(data_ + offset, data, std::min(data_size, array_size - offset));
+                    offset += std::min(data_size, array_size - offset);
+                } while (offset < array_size);
             }
 
             /**
--- a/sparta/sparta/utils/BitArray.hpp
+++ b/sparta/sparta/utils/BitArray.hpp
@@ -219,7 +219,7 @@ namespace sparta {
                 sparta_assert(data_ == nullptr);
 
                 if (data_size > SMALL_OPTIMIZATION_SIZE) {
-                    data_ls_.reserve(data_size);
+                    data_ls_.resize(data_size, 0);
                     data_ = &data_ls_[0];
                 } else {
                     data_ = data_ss_;
--- a/sparta/sparta/utils/BitArray.hpp
+++ b/sparta/sparta/utils/BitArray.hpp
@@ -64,11 +64,7 @@ namespace sparta {
                 }
 
                 initializeStorage_(roundUpTo_(array_size, sizeof(data_type)));
-                size_t offset = 0;
-                do {
-                    std::memcpy(data_ + offset, data, std::min(data_size, array_size - offset));
-                    offset += std::min(data_size, array_size - offset);
-                } while (offset < array_size);
+                std::memcpy(data_, data, std::min(data_size, array_size));
             }
 
             /**
--- a/sparta/test/Register/Register_test.cpp
+++ b/sparta/test/Register/Register_test.cpp
@@ -84,6 +84,9 @@ Register::Definition reg_defs[] = {
                                                                                                                   {"b63_00",  "B", 0,  63,  false} },  {},      nullptr,            Register::INVALID_ID, 0, (const uint8_t*) ALTERNATING_DEFAULT, 0, 0 },
     { 14, "wm_04",    Register::GROUP_NUM_NONE, "",  Register::GROUP_IDX_NONE, "mask spans u64s",           16, { {"b65_13",  "A", 13, 65,  true },
                                                                                                                   {"b67_64",  "B", 64, 67,  false} },  {},      nullptr,            Register::INVALID_ID, 0, (const uint8_t*) ALTERNATING_DEFAULT, 0, 0 },
+
+    // Test very huge register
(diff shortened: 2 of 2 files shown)


2henwei:
@knute-sifive I tried to reproduce the bug on `Register_test` but I can't.

Then I found the problem may be caused by the following code snippet:
https://github.com/sparcians/map/blob/c9c12b24dec0c54ed0decd62a810d1ace7766c34/sparta/sparta/utils/BitArray.hpp#L221-L225

```data_ss_``` was initialized to zero at the beginning so the ```write_mask``` is also correctly initialized to zero. In `Register_test`, large register is assigned exactly 16-bytes, which fits ```SMALL_OPTIMIZATION_SIZE```.

https://github.com/sparcians/map/blob/c9c12b24dec0c54ed0decd62a810d1ace7766c34/sparta/sparta/functional/Register.hpp#L1327-L1333

On the other hand, register size exceeding 16-bytes will be allocated an non-initialized memory space, ```data_ls_```. Its ```write_mask``` may hence have some issues. So I suggested:

1. Initialize the ```data_ls_``` after the memory space is allocated. (So the original commit in this PR doesn't need anymore)
2. Add a "huge" register test (32-bytes) in Register_test.

**Comment by knute-sifive** [comment]
> So I suggested:
> 
>     1. Initialize the `data_ls_` after the memory space is allocated. (So the original commit in this PR doesn't need anymore)
> 
>     2. Add a "huge" register test (32-bytes) in Register_test.
> 
> 
> What do you think?

Sounds like a good plan!

**Comment by 2henwei** [comment]
@knute-sifive, I add a simple huge (256b) register test (basically the same as the large register test). If you directly checkout that commit 7126c5ce8eba9f9fa558014e2f07db7b77045ea3, you would find the following error:

```
Writing to <top.dummy.regs.huge 256 bits cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc 
cc cc cc cc cc cc cc cc cc cc cc cc cc cc>                                                     
 now: <top.dummy.regs.huge 256 bits cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc
 cc cc cc cc ff ff ee fd ff ff ff ff>                                                          
 now: <top.dummy.regs.huge 256 bits cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc
 cc cc cc cc ee ee ee ec ff ff ff ff>                                                          
 now: <top.dummy.regs.huge 256 bits cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc
 cc cc cc cc dd dd ee ec ff ff ff ff>                                                          
 now: <top.dummy.regs.huge 256 bits cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc
 cc cc cc cc dd dd ee ec dd dd ff ff>                                                          
 now: <top.dummy.regs.huge 256 bits cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc
 cc cc cc cc cc dd ee ec dd dd ff ff>                                                          
 now: <top.dummy.regs.huge 256 bits cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc
 cc cc cc cc cc dd ee ec cc dd ff ff>                                                          
Test '(huge->read<uint64_t>())' FAILED on line 956 in file /home/zhenw/athena-devel/map/sparta/
test/Register/Register_test.cpp. Value: '14762217854407147519' should equal '147622178629970821
11'                                                                                            
```

File: sparta/test/Register/Register_test.cpp [hunk 1]
```diff
@@ -929,6 +937,25 @@ int main()
...
+    std::cout << " now: " << huge << std::endl;
+    EXPECT_NOTHROW((huge->write<uint16_t>(0xdddd, 1))); // 3-2
+    std::cout << " now: " << huge << std::endl;
+    EXPECT_NOTHROW((huge->write<uint8_t>(0xcc, 7))); // 7-0 (MSB)
+    std::cout << " now: " << huge << std::endl;
+    EXPECT_NOTHROW((huge->write<uint8_t>(0xcc, 3))); // 7-4
+    std::cout << " now: " << huge << std::endl;
+    std::cout << " have: " << std::hex << huge->read<uint64_t>() << " expect: " << 0xccddeeeeccddffff << std::endl;
+    EXPECT_EQUAL((huge->read<uint64_t>()), 0xccddeeeeccddffff);
+    std::cout << "Huge Register: " << std::endl << huge->renderSubtree(-1, true);
```
knute-sifive, line 957: This testing is great.. a few changes requested:
1. After each write to the given index, add an `EXPECT_EQUAL` to ensure that index was written properly:
```
EXPECT_NOTHROW((huge->write<uint32_t>(0xeeeeeeee, 1))); // 1-0 (MSB)
EXPECT_EQUAL(huge->read<uint32_t>(1). 0xeeeeeeee);
```
1. The `cout` statements aren't too useful if no one looks at them.  :)  Turn them into `EXPECT_EQUAL` statements.

sparta/sparta/utils/BitArray.hpp
```diff
@@ -220,6 +220,7 @@ namespace sparta {
 
                 if (data_size > SMALL_OPTIMIZATION_SIZE) {
                     data_ls_.reserve(data_size);
+                    std::memset(&data_ls_[0], 0, data_size * sizeof(data_type));
```
knute-sifive, line 223: Instead of reserving the memory, resize the vector with the value 0.

2henwei: How could I forget this function? Thanks!

sparta/test/Register/Register_test.cpp
2henwei: Done.

2henwei:
@knute-sifive Thanks for your reviewing and approval! Just a kind reminder that I don't have write access to this repository. (I cant merge it)

PR #165, opened 2020-06-28
Knute/issue 164

### LINKED ISSUE #164: add a public API to sparta::SimulationConfiguration
As a simulator writer I want to access non-sparta configuration files specific to my simulator.  It would be helpful to look for those files on `--arch-search-dir`, but the simulator needs a way to ask sparta for that path information.

klingaard:
closes #164 

Added a few getters to SimulationConfiguration and added some tests.


helios/pipeViewer/scripts/gen_alf/exe_column.cmd
```diff
@@ -0,0 +1,121 @@
+###########################################################################
```
klingaard, line 1: This file accidentally got in... It's ok though.  It was missing initially.

PR #184, opened 2020-08-05
Fix searching for pair transactions in Argos

bdutro:
This PR does some cleanup on transaction_search.cpp and enables searching in Argos for pair_t transactions.


helios/pipeViewer/argos_dumper/DatabaseDump/Database_dumper.cpp
```diff
@@ -26,164 +26,146 @@ namespace pipeViewer
...
+            virtual void foundMemRecord(memoryoperation_t*t)
             {
-                merge_transactions = merge;
-                sort_by_end_time = sort;
+                genericTransactionHandler<memoryoperation_t>(t, printMemOp);
             }
 
+            static void printPairOp(pair_t * p, std::ostream & os){
+                os << "*pair* ";
+                for (uint32_t i = 0; i < p->nameVector.size(); ++i) {
```
klingaard, line 144: Can use a range based for loop

bdutro: Sadly, no. We're iterating over 2 vectors in that loop.

TopChangwatchaiSiFive:
This looks awesome.  Looks like you moved some of the processing from transactiondb.pyx into Reader.hpp, which should speed things up.  But I wanted to understand one thing:  are we effectively throwing away all the pair info when we read transactions from the database?  It appears with `formatPairAsAnnotation` we are allowing everything to be "flattened" so everything will basically be an annotation...is that true?

bdutro:
@TopChangwatchaiSiFive we already have to flatten the pairs into annotation-style strings in order for Argos to handle them properly. This just does the same thing for search purposes. I think this is the most intuitive way to handle it while still providing powerful searching with regexes (unless we want to come up with a SQL-style query language for directly matching on pairs).

PR #194, opened 2020-08-12
fix build fail if map is not a root project of cmake

lihsinyi:
CMAKE_BINARY_DIR is not map if it is not root project.
To replace CMAKE_BINARY_DIR with MAP_BINARY_DIR

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,8 @@ cmake_minimum_required (VERSION 3.14)
 
 project(MAP)
 
+set (MAP_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR})
+
 set (SPARTA_BASE ${CMAKE_CURRENT_SOURCE_DIR}/sparta)
 
 add_subdirectory (sparta)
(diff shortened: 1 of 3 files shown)


knute-sifive:
Hey Hsinyi, does this address the issue of making map a submodule of a project?

knute-sifive:
Just to note, I verified that this change does not affect users who only build the sparta components (and not all of MAP).

lihsinyi:
@knute-sifive Yes, this is because I import map as a submodule in the RTL to pipeline tool.

lihsinyi:
> Just to note, I verified that this change does not affect users who only build the sparta components (and not all of MAP).

What is the exactly target of CPPCHECK_EXCLUDES? It was map binary directory when build all of map but sparta binary directory when build only sparta.

knute-sifive:
Some history on the build of MAP/Sparta...

When the MAP repo was first formed, only the Sparta components had a build target and system.  MAP didn't have the Helios tools yet.  When those tools were added, a build system for _MAP_ was added, which included Sparta.  A developer could still build only Sparta if he/she wanted to. 

But now we need consider having only one build environment with different targets.  Another thing to add to the plate.  

CPPCHECK_EXCLUDES was an attempt to get Sparta CPP check clean.  It's fairly clean, with the exception of a few files which I and cpp check disagree on.  That was its purpose.  Unless it's causing you issues, please ignore it.

PR #195, opened 2020-08-14
Don't destruct data objects in Buffer::erase() or Buffer::clear(),

kevin-heuer:
since the data objects are currently never re-constructed; also,
ensure refcount is decremented when a PayloadDeliveringProxy with a
smart pointer payload is reclaimed for a PhasedPayloadEvent


knute-sifive:
As I suspected, the Buffer_test failed with this change.  Kevin and I chatted about this and really we need Buffer to do what the other resources to -- inplace new/delete.  I'll work on that this week.

knute-sifive:
Hey Kevin, I implement the "proper" fix for this in my branch.  I'll test with SiFive simulators and then put up for review.  I'll close this PR.

PR #196, opened 2020-08-18
Knutel/issue 169 buffer erase

### LINKED ISSUE #169: sparta::Buffer<sparta::SpartaSharedPointer<U>>::erase() should decrement refcount
`sparta::Buffer<sparta::SpartaSharedPointer<U>>::erase()` doesn't decrement the refcount of the `sparta::SpartaSharedPointer<U>` being erased (because `sparta::Buffer<T>` code is rightfully generic and has no idea `T == sparta::SpartaSharedPointer<U>` for `sparta::Buffer<sparta::SpartaSharedPointer<U>>` instances), **which may cause `U` instances to hang around longer than expected**.

Users can work around this by using `*itr = nullptr;` right before `erase(itr);`, but it would be best to handle this automatically in `sparta::Buffer` with template specialization or whatever makes the most sense.

E.g. change `erase()` to `eraseImpl_()`; `sparta::Buffer<T>::erase()` will only call `eraseImpl_()`; `sparta::Buffer<sparta::SpartaSharedPointer<U>>::erase()` will execute `*itr = nullptr;` before calling `eraseImpl_()`.

I haven't been able to get this specific template specialization to work, but this is the general idea.

knute-sifive:
resolves #169 

- `Buffer` now uses placement `new` for allocating its contained objects
- `Buffer` destructs those objects on deallocation (erase)
- Cleared the payload in the PLE if the data type is a pointer
- General class cleanup (trying to use less Boost)


sparta/sparta/events/PhasedPayloadEvent.hpp
```diff
@@ -175,7 +176,9 @@ namespace sparta
         //! reusable.
         void reclaimProxy_(typename ProxyInflightList::iterator & pl_location) {
             sparta_assert(pl_location != inflight_pl_.end());
-            free_pl_[free_idx_++] = *pl_location;
+            if constexpr(MetaStruct::is_any_pointer<DataT>::value) {
+                (*pl_location)->setPayload_(nullptr);
+            }            free_pl_[free_idx_++] = *pl_location;
```
kevin-heuer, line 181: Is `}            free_pl_[free_idx_++] = *pl_location;` two code lines that accidentally got merged into one code line? Or is this intended to be one line of code?

knute-sifive: Copy/paste error.  Thanks for finding.

PR #202, opened 2020-08-31
Buffer bug where internals were not being deallocated on desctruction

knute-sifive:


knute-sifive:
Regression passed locally.  I'll push this fix.

PR #205, opened 2020-09-02
Fix travis

timsnyder:
The cheezy way.  Pin gcc back to where it was working.  Wanted to go ahead and commit this as an example of the syntax for doing it.

### DIFF
--- a/conda.recipe/conda_build_config.yaml
+++ b/conda.recipe/conda_build_config.yaml
@@ -43,7 +43,7 @@ c_compiler:
   - vs2017                     # [win]
 c_compiler_version:            # [unix]
   - 11                         # [osx and arm64]
-  - 10                         # [osx and x86_64]
+  - 9                          # [osx and x86_64]
   - 7.5.0 h09487f9_20          # [linux64 or aarch64]
   - 8                          # [ppc64le or armv7l]
 cxx_compiler:
@@ -52,6 +52,6 @@ cxx_compiler:
   - vs2017                     # [win]
 cxx_compiler_version:          # [unix]
   - 11                         # [osx and arm64]
-  - 10                         # [osx and x86_64]
+  - 9                          # [osx and x86_64]
   - 7.5.0 h09487f9_20          # [linux64 or aarch64]
   - 8                          # [ppc64le or armv7l]
--- a/conda.recipe/conda_build_config.yaml
+++ b/conda.recipe/conda_build_config.yaml
@@ -33,3 +33,25 @@ zip_keys:                       # [osx]
   -                             # [osx]
     - macos_min_version         # [osx]
     - MACOSX_DEPLOYMENT_TARGET  # [osx]
+
+
+# pin linux compiler to old build before the sysroot change because clang 9 wasn't updated
+# to work with latest build of gcc
+c_compiler:
+  - gcc                        # [linux]
+  - clang                      # [osx]
+  - vs2017                     # [win]
+c_compiler_version:            # [unix]
+  - 11                         # [osx and arm64]
+  - 10                         # [osx and x86_64]
+  - 7.5.0 h09487f9_20          # [linux64 or aarch64]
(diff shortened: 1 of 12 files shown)


conda.recipe/conda_build_config.yaml
```diff
@@ -33,3 +33,25 @@ zip_keys:                       # [osx]
...
+# pin linux compiler to old build before the sysroot change because clang 9 wasn't updated
+# to work with latest build of gcc
+c_compiler:
+  - gcc                        # [linux]
+  - clang                      # [osx]
+  - vs2017                     # [win]
+c_compiler_version:            # [unix]
+  - 11                         # [osx and arm64]
+  - 10                         # [osx and x86_64]
+  - 7.5.0 h09487f9_20          # [linux64 or aarch64]
```
timsnyder, line 47: If you want to lock down the gcc compiler to a specific version, you need to do this.  I'm also now realizing why conda is having trouble solving our environment on Mac.  The recipe asks for clang 10 and clang 9.  I'm going to let the CI run once to make sure before I push more commits to fix that.

knute-sifive (approved):

conda.recipe/conda_build_config.yaml
knute-sifive: Looks like it's indeed asking for both.

knute-sifive:
> Commits I added this morning should make the Mac builds work again as well. I'll click the merge button if both are passing and then open another PR that upgrades to using clang 10 since you mentioned the code supports it.

Thank you!

timsnyder:
```
cd $SRC_DIR/release/helios/pipeViewer/transactionsearch && $BUILD_PREFIX/bin/cmake -E cmake_link_script CMakeFiles/transactionsearch.dir/link.txt --verbose=1
$BUILD_PREFIX/bin/x86_64-apple-darwin13.4.0-clang++ -march=core2 -mtune=haswell -mssse3 -ftree-vectorize -fPIC -fPIE -fstack-protector-strong -O2 -pipe -stdlib=libc++ -fvisibility-inlines-hidden -std=c++14 -fmessage-length=0 -isystem $PREFIX/include -fdebug-prefix-map=$SRC_DIR=/usr/local/src/conda/map-untagged -fdebug-prefix-map=$PREFIX=/usr/local/src/conda-prefix -O3 -DNDEBUG -isysroot /Applications/Xcode_11.6.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.13.sdk -mmacosx-version-min=10.13 -Wl,-search_paths_first -Wl,-headerpad_max_install_names -Wl,-pie -Wl,-headerpad_max_install_names -Wl,-dead_strip_dylibs -Wl,-rpath,$PREFIX/lib -L$PREFIX/lib CMakeFiles/transactionsearch.dir/src/transaction_search.cpp.o -o transactionsearch  ../../../sparta/libsparta.a ../../../sparta/simdb/libsimdb.a $PREFIX/lib/libhdf5.dylib -lsqlite3 $PREFIX/lib/libyaml-cpp.0.6.3.dylib -lz -lpthread -lrt $PREFIX/lib/libboost_date_time.dylib $PREFIX/lib/libboost_filesystem.dylib $PREFIX/lib/libboost_iostreams.dylib $PREFIX/lib/libboost_serialization.dylib $PREFIX/lib/libboost_timer.dylib $PREFIX/lib/libboost_program_options.dylib $PREFIX/lib/libboost_chrono.dylib 
ld: library not found for -lrt
clang-9: error: linker command failed with exit code 1 (use -v to see invocation)
```

Now I'm remembering the problem with just adding `-lrt` all the time.  You have to do it only on Linux there isn't such a think on OSX.  I'm going to tweak the `sparta-config.cmake` to only add `rt` when we're not on Apple to make this work for now.

knute-sifive:
Woo hoo!

PR #206, opened 2020-09-03
Use clang10 in ci

timsnyder:
update conda.recipe to use clang 10 in the CI.  After updating to clang 10, we no longer need to lock down the gcc build.


timsnyder:
Hey @klingaard , the error in the Mac build seems like something you might legitimately want to change:
```
n file included from /Users/runner/miniforge3/conda-bld/map_1599160748327/work/sparta/simdb/test/Utils/Utils_test.cpp:7:
/Users/runner/miniforge3/conda-bld/map_1599160748327/work/sparta/./simdb/test/SimDBTester.hpp:113:33: error: loop variable 's' of type 'const std::__1::basic_string<char>' creates a copy from type 'const std::__1::basic_string<char>' [-Werror,-Wrange-loop-construct]
                for (const auto s : methods_reached_) {
                                ^
/Users/runner/miniforge3/conda-bld/map_1599160748327/work/sparta/./simdb/test/SimDBTester.hpp:113:22: note: use reference type 'const std::__1::basic_string<char> &' to prevent copying
                for (const auto s : methods_reached_) {
                     ^~~~~~~~~~~~~~
                                &
```
there's also another one of the same type of thing that should be `const auto &` instead of `const auto`.  Is that something you want me to add to this PR?

timsnyder:
Failure on Linux is the same test code that should probably be using `const auto &`.  I'm going to go ahead and push another commit here to clean that up.

knute-sifive (approved):

Issue #200, opened 2020-08-28
Argos segfault when running with wxPython 4.1.0


criusx:
Running under Ubuntu 20.04, though not strictly necessary, pip3 install wxPython==4.1.0, run the CoreExample, then argos.py and you will see a segfault when trying to call DrawItem(). This happens at core.pyx:550 when tryint to update the annotation. Workaround is to revert to a lower wxPython (e.g. pip3 install wxPython==4.0.7). Hope this saves somebody trouble since 4.1.0 is the latest and default version now. Nothing in the 4.1.0 release notes seems to indicate this would happen. Ideally argos can work with all versions of wxPython Phoenix(>v4) and will exit with an error message if an unsupported version is loaded. This isn't urgent, but someone else familiar with the code may be able to fix it easily.

`#0  0x0000000000000003 in  ()                                       
#1  0x00007ffff6009bd5 in wxGCDCImpl::DoDrawBitmap(wxBitmap const&, int, int, bool) () at /usr/local/lib/python3.8/dist-packages/wx/libwx_gtk3u_core-3.1.so.4                                                                                                                             
#2  0x00007fffef5c7d5f in wxDC::DrawText(wxString const&, int, int) (this=0x2271490, y=20, x=421, text=...) at /usr/include/wx-3.0/wx/dc.h:1218
#3  __pyx_pf_4core_8Renderer_20drawInfoRectangle                                                                                                                                                                                                                                          
    (__pyx_v_short_format=<optimized out>, __pyx_v_schedule_settings=<optimized out>, __pyx_v_clip_x=<optimized out>, __pyx_v_auto_color=<optimized out>, __pyx_v_content_type=<optimized out>, __pyx_v_missing_needed_loc=<optimized out>, __pyx_v_annotation=<optimized out>, __pyx_v_re
ct=<optimized out>, __pyx_v_canvas=<optimized out>, __pyx_v_dc=<optimized out>, __pyx_v_element=<optimized out>, __pyx_v_tick=<optimized out>, __pyx_v_self=0x7fffeec27c70) at /home/osboxes/map/helios/pipeViewer/pipe_view/core/src/core.cpp:8219                                       
[...]

PR #211, opened 2020-10-06
Argos compatibility fixes for GTK3 and wxPython 4.1.0

bdutro:
This PR makes Argos compatible with GTK3 and the latest version of wxPython (4.1.0). It also includes a workaround for #200 (see https://github.com/sparcians/map/issues/200#issuecomment-703932506 for details).

I've tested this locally against GTK2, GTK3, and OS X, but it would be nice if someone else could take a test drive on their setup and make sure things still work.


bdutro:
I'm not sure why this test is failing. Regression passes on my machine.
`179/179 Test #164: dynamic_model_pipeline_csv_headers_in_db ..................................***Failed   12.48 sec`

knute-sifive (approved):

PR #222, opened 2021-01-27
wxPython 4.1.0 fixes

bdutro:
Fixes for crashes/rendering issues observed when running with wxPython 4.1.0:
- Database and layout chooser dialogs caused crashes due to invalid sizer settings
- Buttons with BU_EXACTFIT style were not drawing their borders


klingaard:
Looks like -lrt isn't available on Mac

klingaard (approved):

bdutro:
That was weird. I don't know how those extra commits ended up in the branch.

PR #223, opened 2021-02-03
Knutel/issue 152 improve register write speed

knute-sifive:
This PR introduces a more optimized register bits manipulation class.  The class removes the need for dynamic allocations and copies on register reads as well as providing optimized paths for bit-wise operations.  I'm sure there are cleaner things to be done, but the register performance test went from taking 20 seconds to 5.


helios/plato/django/server/utilityEndpoint.py
```diff
@@ -91,6 +91,9 @@ async def saveLayout(self, jsonData, returnValue):
                         returnValue["error"] = ""
                 except IntegrityError as e:
                     returnValue["error"] = f"not created, {str(e)}"
+                except:
+                        returnValue["result"] = "success"
+                        returnValue["error"] = ""
```
knute-sifive, line 96: Oh yeah, this is a bug fix to the Django server stuff.  Wondered where it went.  :laughing:

sparta/sparta/functional/RegisterBits.hpp
```diff
@@ -0,0 +1,704 @@
...
+            return final_value;
+        }
+
+        // Copy the remote register data locally.
+        void convert_() {
+            if(nullptr == local_data_) {
+                local_data_ = local_storage_.data();
+                ::memset(local_data_, 0, local_storage_.size());
+                ::memcpy(local_data_, remote_data_, num_bytes_);
+                remote_data_ = local_data_;
```
knute-sifive, line 70: Believe it or not, doing this `memset` here is _faster_ than doing `local_data_ = {0}` initialization at the member declaration.

sparta/sparta/functional/RegisterBits.hpp
```diff
@@ -0,0 +1,704 @@
...
+         * \param data The data to write to the lower portion of memory
+         *
+         * The data is copied
+         */
+        template<class DataT>
+        RegisterBits(const uint64_t num_bytes, const DataT & data) :
+            local_storage_(),
+            local_data_(local_storage_.data()),
+            remote_data_(local_data_),
+            num_bytes_(num_bytes)
```
knute-sifive, line 101: Clang bug!  You _have to_ initialize these variables in the member list or clang will barf.  Only happens on template class constructors.  It's been reported.

klingaard (approved):

PR #225, opened 2021-03-04
Let events can be scheduled before/after pipeline updates

2henwei:
To forward data through the pipe, Sparta Pipeline Class schedules a "pipeline update" event in the Update phase to do it.

However, the data at the stage N in Cycle T always comes from the stage N-1 in Cycle T-1. This makes us hard to insert a new data into the specific stage half through the pipeline.

The newly added method, setConsumerForPipelineUpdate, enables us to schedule an event just after pipeline updates. After the pipeline forwards the data, now we have flexibility to modify data at the specific stage before collection/tick phases. But be noted that the handler of stage might need to be scheduled by user because the modified stage can be originally invalid in this cycle. Pipeline class wouldn't schedule an event for an invalid stage.

### DIFF
--- a/sparta/sparta/resources/Pipeline.hpp
+++ b/sparta/sparta/resources/Pipeline.hpp
@@ -400,6 +400,38 @@ namespace sparta
         }
 
         /*!
+         * \brief Specify producer event for the pipeline update event
+         * \param ev_handler The producer event handler
+         *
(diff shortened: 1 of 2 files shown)


klingaard (approved):
Nice addition to the pipeline class.  Thanks!

PR #232, opened 2021-04-06
Fixed the issue with two diff allocators being used for common SPP

### LINKED ISSUE #231: SpartaSharedPointer's of derived types, using different allocators will segfault
Thought we had some good testing for this, but a use case just popped up where two SpartaSharedPointer (SPP) types (one being a derivative of the other) have two different allocators causes a crash (seg fault) when the base type SSP is released.
Rough example:
```
{
   SSP<Base> base; 
   SSP<Derived> der = allocator::allocate_sparta_shared_pointer<Derived>(...);
   base = der;
   der.reset();
}
// crash
```

knute-sifive:
closes #231

### DIFF
--- a/sparta/sparta/utils/SpartaSharedPointer.hpp
+++ b/sparta/sparta/utils/SpartaSharedPointer.hpp
@@ -17,6 +17,38 @@
 
 namespace sparta
 {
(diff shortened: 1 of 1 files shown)


kathlenehurt-sifive (approved):
LGTM.

PR #233, opened 2021-04-06
Knutel/issue 209 no extentions in final

### LINKED ISSUE #209: --write-final-config doesn't show extension values
Subject says it all

knute-sifive:
Extensions are now in the final config

@MichaelSchoenfelder-SiV 

closes #209


MichaelSchoenfelder-SiV (approved):
The following sequence works (after adjusting locations of yamls):
```
 1184  ./dabble --arch mallard/u84_standard.yaml --write-final-config u84_standard_final.yaml --no-run
 1185  ./dabble --arch mallard/aquarius.yaml --write-final-config aquarius_final.yaml --no-run
 1186  ./dabble --arch u84_standard_final.yaml --arch aquarius_final.yaml --write-final-config final_final.yaml --no-run
```
I am concerned that the final config format includes lines with `---` and `...`.  I don't think that is valid yaml format, but apparently the dabble yaml parser doesn't care.  And those lines were there before, so have nothing to do with this change.

PR #235, opened 2021-04-07
Argos: Add persistent settings file and variable font sizes

### LINKED ISSUE #234: In Argos change font size of tooltips
@benoy-sifive reported this.  On his Mac, the tooltips font is too small.

The main Argos window can be resized using `Ctrl`+scroll (`Cmd`+scroll on Mac), or using `Ctrl` + Plus, Minus, or Zero keys.

However, this doesn't change the font size of tooltips.

Currently, the tooltip font size is set here:  https://github.com/sparcians/map/blob/4026a37409f95761df34d6140c47630478bc09d0/helios/pipeViewer/pipe_view/gui/hover_preview.py#L46

Two possible solutions:
1. Add persistent settings that users can set
2. Tooltip should zoom in/out the same as the main Argos window (not sure how this would be done)

bdutro:
This PR adds a new menu option (Settings...) to the View menu that allows changing the default font sizes for the layout canvas and hover window. The font sizes are stored in a persistent settings file in the OS-appropriate user home directory.

Fixes #234

### DIFF
--- a/helios/pipeViewer/pipe_view/argos.py
+++ b/helios/pipeViewer/pipe_view/argos.py
@@ -12,6 +12,7 @@ import os
 import logging
 import pstats
 from logging import info, debug, error, warn
+from model.settings import ArgosSettings
 from model.utilities import LogFormatter
 import cProfile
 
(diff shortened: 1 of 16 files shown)


TopChangwatchaiSiFive (approved):

knute-sifive (approved):

benoy-sifive:
Confirm the latest branch works on MacOS

PR #239, opened 2021-04-13
Added a basic histogram class with user-provided buckets

eabsi5:


sparta/sparta/statistics/BasicHistogram.hpp
```diff
@@ -0,0 +1,63 @@
...
+        // create a final bucket for everything beyond the last value
+        std::ostringstream os_name;
+        os_name << name << "_of";
+        std::ostringstream os_desc;
+        os_desc << desc << " overflow";
+        ctrs_.push_back(sparta::Counter(&sset, os_name.str(), os_desc.str(), sparta::Counter::COUNT_NORMAL));
+    }
+
+    /// increment the bucket corresponding to 'val'
+    void addValue(int64_t val)
```
klingaard, line 52: What happens if I pass in `-1`?

klingaard (changes requested):
Missing testing.

sparta/sparta/statistics/BasicHistogram.hpp
```diff
@@ -0,0 +1,63 @@
...
+    {
+        auto bucket = std::lower_bound(bucket_vals_.begin(), bucket_vals_.end(), val);
+        auto off = bucket - bucket_vals_.begin();
```
klingaard, line 55: Assumes that bucket is valid.

sparta/sparta/statistics/BasicHistogram.hpp
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * \file BasicHistogram.hpp
+ * \brief A simple histogram with programmable ranges, using sparta::Counters
+ */
+#pragma once
+
+#include "sparta/statistics/Counter.hpp"
+
+/// Histogram with programmable buckets
+class BasicHistogram
```
klingaard, line 12: I think this class could be extended easily to include different bucket types (not just `vector of ints`  as well as a user defined comparator function that can be passed to the `lower_bound` method.

sparta/sparta/statistics/BasicHistogram.hpp
eabsi5: If it's bucket_vals_.end() then you get bucket_vals_.size() which corresponds to the overflow.

sparta/sparta/statistics/BasicHistogram.hpp
eabsi5: You can potentially define negative buckets, if so, it will be sorted into the appropriate bucket. Otherwise, it goes into the first bucket.

klingaard: So, if I pass in `-1` to a group of buckets 0-100, 100-200, 200-300, you'd expect it to be charged to bucket 0-100?  Does that seem right?

eabsi5:
There isn't much that can be tested, but this will at least test construction and destruction (along with addValue)

sparta/sparta/statistics/BasicHistogram.hpp
eabsi5: I don't see how the class can know what are the expectations of the input. If -1 is invalid, there should be an assert in the user code. If negative numbers want to be binned separately, there should be a leading bucket of -1...

eabsi5:
Let me switch from lower_bound to upper_bound. I think that will be more intuitive for 0 and negative buckets.

sparta/sparta/statistics/BasicHistogram.hpp
klingaard: The point of a framework is to not force the user to perform his/her own error checking, but to ensure the user can find issues quickly if they do something wrong.  Clearly, -1 (as well as something like 500) is not a valid input and I think this class should ensure that the values given to it should cleanly fall into a bucket that it has created (i.e. assert).  Either that, or enhance the class to allow for "out of range" values and create buckets for "beyond lower" and "beyond upper."  You have the list of valid ranges, it's not hard to derive what the upper/lower limits are.

klingaard:
> Let me switch from lower_bound to upper_bound. I think that will be more intuitive for 0 and negative buckets.

Cool.  And whilst you're at it, add testing for out-of-order vector and templatize the class to take other types of buckets other than signed ints.  (I'd like to use it for `class enum` for example).

eabsi5:
I think that should work...

eabsi5:
Will we be able to rebase on merge?

klingaard (approved):
Looking good.  Might be worth checking to make sure the buckets got the right counts in the right places.  No testing for that.

klingaard:
> Will we be able to rebase on merge?

Yes

eabsi5:
Yeah, unsigned gets -1, so the loop wasn't running...

PR #240, opened 2021-04-20
Fix spurious transaction value changes in Argos

bdutro:
This PR fixes a bug in Argos collection that caused collectors to grab the latest string data for a location every time that location was written to a database. Unfortunately, this often resulted in it getting the string data for the _next_ transaction instead of the current one.

A common example of this bug is an instruction in a sparta::Queue that appears to suddenly change its disassembly string several cycles before the next instruction actually enters that entry.

### DIFF
--- a/sparta/sparta/collection/Collectable.hpp
+++ b/sparta/sparta/collection/Collectable.hpp
@@ -574,7 +574,7 @@ namespace sparta{
                 updateLastRecord_();
 
                 // If the new Value pairs are not empty and current record is closed, we start a new record.
-                if(!last_record_values_.empty() && record_closed_){
+                if(!(last_record_values_.empty() && last_string_values_.empty()) && record_closed_){
                     startNewRecord_();
                     record_closed_ = false;
                 }
@@ -602,7 +602,7 @@ namespace sparta{
                 updateLastRecord_();
 
                 // If the new Value pairs are not empty and current record is closed, we start a new record.
-                if(!last_record_values_.empty() && record_closed_){
+                if(!(last_record_values_.empty() && last_string_values_.empty()) && record_closed_){
                     startNewRecord_();
                     record_closed_ = false;
                 }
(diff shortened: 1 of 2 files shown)


BrianGraysonSiV:
I don't see any comments that have been added. Is the code self-documenting enough that you don't want to add any? I haven't looked at these guts, so I can't tell at a glance whether your changes are obvious, or magical!

bdutro:
> I don't see any comments that have been added. Is the code self-documenting enough that you don't want to add any? I haven't looked at these guts, so I can't tell at a glance whether your changes are obvious, or magical!

Thanks, I took another look at the code and realized it could be simplified. I added some comments too!

sparta/sparta/collection/Collectable.hpp
```diff
@@ -758,7 +758,35 @@ namespace sparta{
             //! \brief Fill it with the new Name Value pairs from
             //  this cycle of Collection.
             void updateLastRecord_(){
-                last_record_values_ = getDataVector();
+                // These fields only need to be set once - they define the format of the collectable
+                // and remain constant for the duration of the simulation
+                if(SPARTA_EXPECT_FALSE(argos_record_.nameVector.empty())) {
+                    argos_record_.nameVector = getNameStrings();
+                    argos_record_.length = argos_record_.nameVector.size();
+                    has_display_id_field_ = argos_record_.nameVector[0] == "DID";
```
BrianGraysonSiV, line 766: I would add parentheses around the == expression. Yes, it's redundant, but it aids readability so that one doesn't scan it and think has_display_id_field_ = argos_record_.nameVector[0] at a rapid glance.

bdutro: Fixed.

klingaard (approved):

sparta/sparta/collection/Collectable.hpp
```diff
@@ -758,7 +758,35 @@ namespace sparta{
             //! \brief Fill it with the new Name Value pairs from
             //  this cycle of Collection.
             void updateLastRecord_(){
-                last_record_values_ = getDataVector();
+                // These fields only need to be set once - they define the format of the collectable
+                // and remain constant for the duration of the simulation
```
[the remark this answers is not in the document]
bdutro: They can, and they are now moved.

PR #242, opened 2021-04-21
Fix for Argos display bug where a transaction would sometimes extend for an extra cycle

bdutro:
This PR fixes an Argos display bug that caused transactions to occasionally extend for an extra cycle if they ended at a heartbeat interval.

### Background
Whenever a heartbeat event occurs, any open transactions are marked as "continuing," which tells Argos that they should be extended past the heartbeat boundary when they are rendered. Transactions are marked closed at collection time whenever the underlying data changes.

### Root Cause
The PipelineCollector heartbeat event is scheduled during the Collection phase without any precedence set, which means it can occur in any order with respect to other events in that phase. The Collection phase is also the default scheduler phase for all pipeline collection events. Any locations that have their collection events scheduled before the heartbeat event wouldn't be affected by this bug, while any locations after the event could be erroneously marked as "continuing" (see images below for an example of how this affects Argos' rendering).

#### Examples of Bug (left) and Fix (right)
bug fix

### Fix
This fix ensures that *all* pipeline collection occurs before the heartbeat event by moving the heartbeat event to the PostTick phase and adding a Global Ordering Point (GOP) between all other events in PostTick and the heartbeat event. This was necessary because it's possible for collection to happen in any scheduler phase.

event_diagram

### DIFF
--- a/sparta/sparta/collection/PipelineCollector.hpp
+++ b/sparta/sparta/collection/PipelineCollector.hpp
@@ -18,6 +18,7 @@
 #include "sparta/collection/Collector.hpp"
 #include "sparta/events/EventSet.hpp"
 #include "sparta/events/UniqueEvent.hpp"
+#include "sparta/events/GlobalOrderingPoint.hpp"
 #include "sparta/kernel/Scheduler.hpp"
 
 #include "sparta/pipeViewer/Outputter.hpp"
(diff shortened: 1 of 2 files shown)


klingaard:
Hmm...  I wonder if you just had the `PostTick` GOP precced the `CollectionHeartbeat` GOP, disconnect the heartbeat event from the `PostTick` GOP and connect it to the ``CollectionHearbeat` event...
```
DAG* dag = scheduler_->getDAG(); // Get a handle to the DAG
sparta_assert(dag);
Vertex* heartbeat_gop = dag->getGOPoint("CollectionHeartbeat"); // Create the CollectionHeartbeat GOP
sparta_assert(heartbeat_gop);
Vertex* post_tick = dag->getGOPoint("PostTick"); // Get a handle to the PostTick GOP
sparta_assert(post_tick);

// Link post -> heartbeat
dag->link(post_tick, heartbeat_gop);

dag->unlink(ev_heartbeat_.getVertex(), post_tick); // Undo the ev_heartbeat_ >> post_tick link
dag->link(ev_heartbeat_.getVertex(), heartbeat_gop); // Hook up heartbeat to heartbeat GOP
```

and remove the code in Scheduleable.cpp

bdutro:
Even better, I can get rid of the new GOP completely and just relink the PostTick GOP to precede ev_heartbeat_.

klingaard (approved):

PR #255, opened 2021-07-11
Added param. to suppress events of stalled stages

2henwei:
Literally, "stall" should just represent pipeline stage forwarding stops. Data are still kept in those stalled stages. In some scenarios, modelers may still want registered events to be triggered to handle data.


klingaard (approved):
I like it!  And thank you for adding testing.

PR #256, opened 2021-07-15
Added support on Pipeline to trigger payload event

2henwei:
Added a new template parameter of Pipeline, EventT, to enhance Pipeline event system.

Copying the description from codes:
> Template parameter DataT specifies the type of data flowing through pipeline stages, and
EventT can be specified by 2 Event types: PhasedUniqueEvent (default) and PhasedPayloadEvent<DataT>.
The difference between both types is related to what kind of SpartaHandler modelers are going
to register at stages. With default PhasedUniqueEvent, you can regsiter a SpartaHandler
with no data; or, with PhasedPayloadEvent<DataT>, you can register a SpartaHandler with
data of type DataT. Pipeline will prepare payload and pass data of the stage to every handler.

I was surprised that the Pipeline implementation had been using EventSet pointed to NULL.

However, PayloadEvent needs an available EventSet. So I added a new parameter in constructor to set it up.

Also a dummy EventSet variable is added to keep the backward compatibility, making codes a little tricky ...

### DIFF
--- a/sparta/sparta/resources/Pipeline.hpp
+++ b/sparta/sparta/resources/Pipeline.hpp
@@ -23,6 +23,7 @@
 #include "sparta/resources/Pipe.hpp"
 #include "sparta/events/Scheduleable.hpp"
 #include "sparta/events/PhasedUniqueEvent.hpp"
+#include "sparta/events/PhasedPayloadEvent.hpp"
 #include "sparta/kernel/SpartaHandler.hpp"
 
 namespace sparta
@@ -38,6 +39,13 @@ namespace sparta
      * It contains a sparta::Pipe, and couples event-scheduling (i.e. control-flow path), with the
      * data-movement (i.e. data-flow path) provided by sparta::Pipe.
      *
+     * Template parameter DataT specifies the type of data flowing through pipeline stages, and
+     * EventT can be specified by 2 Event types: PhasedUniqueEvent (default) and PhasedPayloadEvent<DataT>.
+     * The difference between both types is related to what kind of SpartaHandler modelers are going
+     * to register at stages. With default PhasedUniqueEvent, you can regsiter a SpartaHandler
+     * with no data; or, with PhasedPayloadEvent<DataT>, you can register a SpartaHandler with
+     * data of type DataT. Pipeline will prepare payload and pass data of the stage to every handler.
+     *
      * The sparta::Pipeline is able to provide modelers with the following design capability:
      * -# Register event handler(s) for designated pipeline stage(s), sparse stage handler
      *    registration is supported.
@@ -49,15 +57,15 @@ namespace sparta
      * -# Perform manual or automatic pipeline update (i.e. forward progression). The registered
      *    pipeline stage handlers are called under-the-hood whenever valid pipeline data arrives.
      */
-    template <typename DataT>
+    template <typename DataT, typename EventT=PhasedUniqueEvent>
     class Pipeline
     {
     public:
         using size_type = uint32_t;
         using value_type = DataT;
-        using EventHandle = std::unique_ptr<PhasedUniqueEvent>;
+        using EventHandle = std::unique_ptr<EventT>;
         using EventHandleList = std::list<EventHandle>;
-        using EventList = std::list<PhasedUniqueEvent*>;
+        using EventList = std::list<EventT*>;
         using EventMatrix = std::array<EventList, NUM_SCHEDULING_PHASES>;
 
         /*!
@@ -71,7 +79,7 @@ namespace sparta
             NUM_OF_PRECEDENCE
         };
 
-        template<typename DataT2>
+        template<typename DataT2, typename EventT2>
         friend class Pipeline;
 
         /*!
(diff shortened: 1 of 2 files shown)


sparta/sparta/resources/Pipeline.hpp
```diff
@@ -218,24 +226,33 @@ namespace sparta
...
             ev_pipeline_update_
-                (&es_, name + "_update_event", CREATE_SPARTA_HANDLER(Pipeline, internalUpdate_), 1),
+                (es_, name + "_update_event", CREATE_SPARTA_HANDLER(Pipeline, internalUpdate_), 1),
             num_stages_(num_stages)
         {
+            // Only support the following Event types:
+            // 1. PhasedUniqueEvent
+            // 2. PhasedPayloadEvent<DataT>
+            if constexpr (!std::is_same_v<EventT, PhasedUniqueEvent> && !std::is_same_v<EventT, PhasedPayloadEvent<DataT>>) {
+                sparta_assert(false, "Pipeline is templated a unsupported Event type");
```
knute-sifive, line 254: You can make this a `static_assert` actually.

sparta/sparta/resources/Pipeline.hpp
```diff
@@ -49,15 +57,15 @@ namespace sparta
      * -# Perform manual or automatic pipeline update (i.e. forward progression). The registered
      *    pipeline stage handlers are called under-the-hood whenever valid pipeline data arrives.
      */
-    template <typename DataT>
+    template <typename DataT, typename EventT=PhasedUniqueEvent>
```
knute-sifive, line 60: You might need to include the DataT in the EventT event:
```
typename EventT=PhasedUniqueEvent<DataT>
```

sparta/test/Pipeline/Pipeline_test.cpp
```diff
@@ -288,6 +292,8 @@ int main ()
 
     sparta::Pipeline<uint64_t> examplePipeline7("mySeventhSpartaPipeline", 2, root_clk.get());
 
+    sparta::Pipeline<uint64_t, sparta::PhasedPayloadEvent<uint64_t>> examplePipeline8(&es, "myEighthSpartaPipeline", 3, root_clk.get());
```
knute-sifive, line 295: Curious, did you attempt to pass and event type whose template argument != DataT?  What would happen here?
```
sparta::Pipeline<uint64_t, sparta::PhasedPayloadEvent<std::string>>
```

sparta/sparta/resources/Pipeline.hpp
```diff
@@ -38,6 +39,13 @@ namespace sparta
      * It contains a sparta::Pipe, and couples event-scheduling (i.e. control-flow path), with the
      * data-movement (i.e. data-flow path) provided by sparta::Pipe.
      *
+     * Template parameter DataT specifies the type of data flowing through pipeline stages, and
+     * EventT can be specified by 2 Event types: PhasedUniqueEvent (default) and PhasedPayloadEvent<DataT>.
+     * The difference between both types is related to what kind of SpartaHandler modelers are going
+     * to register at stages. With default PhasedUniqueEvent, you can regsiter a SpartaHandler
```
knute-sifive, line 45: register is misspelled (regsiter)

sparta/sparta/resources/Pipeline.hpp
```diff
@@ -266,22 +296,39 @@ namespace sparta
 
             // Create a new stage event handler, and add it to its event list
             auto & event_list = event_list_at_stage_[id];
-            event_list.emplace_back(new sparta::UniqueEvent<sched_phase>(&es_,
-                                        "uev_stage_" + std::to_string(id) + "_" + std::to_string(event_list_at_stage_[id].size()),
-                                        handler));
+            if constexpr (std::is_same_v<EventT, PhasedPayloadEvent<DataT>>) {
+                event_list.emplace_back(new EventT(es_,
```
knute-sifive, line 300: Add an assert to make sure `handler.argCount() == 1`

knute-sifive (changes requested):
A few changes here and there, but looks good

knute-sifive:
> I was surprised that the Pipeline implementation had been using EventSet pointed to NULL.

Actually, it's not a null EventSet.  It's a concrete object, whose parent is nullptr.

sparta/sparta/resources/Pipeline.hpp
2henwei: But PhasedUniqueEvent doesn't have template parameter?

sparta/test/Pipeline/Pipeline_test.cpp
2henwei: I didn't try it. But I think it probably work until the handler is actually called by Pipeline. In you example, PayloadEvent is expecting std::string but Pipeline feeds DataT ...

Eventually I have an assertion to make sure PhasedPayloadEvent uses the template argument DataT.

sparta/sparta/resources/Pipeline.hpp
2henwei: Thanks!

sparta/sparta/resources/Pipeline.hpp
2henwei: Nice!

2henwei:
> 
> 
> > I was surprised that the Pipeline implementation had been using EventSet pointed to NULL.
> 
> Actually, it's not a null EventSet. It's a concrete object, whose parent is nullptr.

I guess my next question is why PayloadEvent is not scheduleable itself ...

sparta/sparta/resources/Pipeline.hpp
knute-sifive: Right!  Duh.

sparta/test/Pipeline/Pipeline_test.cpp
knute-sifive: Actually, I just tried it -- it successfully fails with the new static_assert you just added.  😁 
```
.././sparta/resources/Pipeline.hpp:253:13: error: static_assert failed due to requirement 'std::is_same_v<sparta::PhasedPayloadEvent<int>, sparta::PhasedUniqueEvent> || std::is_same_v<sparta::PhasedPayloadEvent<int>, sparta::PhasedPayloadEvent<unsigned long> >' "Pipeline is templated a unsupported Event type. Supported Event types: {PhasedUniqueEvent, PhasedPayloaodEvent<DataT>}"
```
when I did this:
```
sparta::Pipeline<uint64_t, sparta::PhasedPayloadEvent<int>> examplePipeline8(&es, "myEighthSpartaPipeline", 3, root_clk.get());
```

knute-sifive: However, the error message should look more like this:
"Error: Pipeline is templated on a unsupported Event type. Supported Event types: UniqueEvent, PayloaodEvent<DataT> (where DataT == DataT of the Pipeline). "

knute-sifive:
> I guess my next question is why PayloadEvent is not scheduleable itself ...

Good design-related question.  😄   I struggled with that for a while, but a use case popped up that made the direction clear.  

If a modeler wanted to create a generic unit that takes a Scheduleable to be scheduled at some later time, the modeler would be "stuck" needing to know if it were a PayloadEvent or a simple Event type (Unique, SingleCycle, etc).  

Additional to that, the modeler would also be stuck needing to know the payload to schedule.  This ... sucked.  So, I created a proxy scheduleable that was generated from the PayloadEvent -- made it look like a simple Event type.   That way I could write generic code that looks like this:
```
void Foo::scheduleEvent(sparta::Scheduleable * sched) {
    sched->schedule();
}
```

The PayloadEvent is really just a factory, creating Scheduleables with a payload.  That way I could wrap a payload in a Scheduleable and pass it as a generic object to a downstream component that had no idea what type of Scheduleable it was.  But the "cost" of this is that PLE is not a directly scheduable object.

knute-sifive (approved):

PR #257, opened 2021-07-19
Scheduler init tick0

robbellphd:
Scheduler init tick on tick 0.

### DIFF
--- a/helios/pipeViewer/pipe_view/model/clock_manager.py
+++ b/helios/pipeViewer/pipe_view/model/clock_manager.py
@@ -55,8 +55,8 @@ class ClockManager:
         #  differently
         def LocalToHypercycle(self, local_cycle):
             assert local_cycle is not None
-            # Absolute cycle counts begin at 1. Subtract 1 before scaling into ticks
-            return int(math.floor((local_cycle - 1) * self.__hc_tick_period))
+            # Absolute cycle counts begin at 0.
+            return int(math.floor((local_cycle) * self.__hc_tick_period))
 
         ## Convert a cycle in terms of hypercycle ticks into number of cycles on
         #  this clock domain.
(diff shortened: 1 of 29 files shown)


knute-sifive:
So I'm going to hold off on this change until our internal tools are tested with it.  Might consider putting it in a local branch...  I'll think about it.  Until then, I'll (or you can) keep it up to date with latest master changes.

robbellphd:
Will do. Feel free to update it, I will also if I see changes to master.

knute-sifive:
We're going to start looking at this next week.  @bdutro-sv had a few Argos changes that need to be made along with this.

knute-sifive:
We got past some major hurdles with this change.  Things are looking good -- one test is still failing.  We're looking at it.

knute-sifive:
Hey Rob, this patch will fix argos as well:
```
diff --git a/helios/pipeViewer/pipe_view/model/clock_manager.py b/helios/pipeViewer/pipe_view/model/clock_manager.py
index 21b933215..a4f232777 100644
--- a/helios/pipeViewer/pipe_view/model/clock_manager.py
+++ b/helios/pipeViewer/pipe_view/model/clock_manager.py
@@ -56,7 +56,7 @@ class ClockManager:
         def LocalToHypercycle(self, local_cycle):
             assert local_cycle is not None
             # Absolute cycle counts begin at 1. Subtract 1 before scaling into ticks
-            return int(math.floor((local_cycle - 1) * self.__hc_tick_period))
+            return int(math.floor((local_cycle) * self.__hc_tick_period))
 
         ## Convert a cycle in terms of hypercycle ticks into number of cycles on
         #  this clock domain.
@@ -67,19 +67,19 @@ class ClockManager:
         #  @return integer number of cycles on this clock domain
         def HypercycleToLocal(self, hc):
             assert hc is not None
-            # Absolute cycle counts begin at 1. Add 1 after scaling into ticks
-            return 1 + int(math.floor(hc / self.__hc_tick_period))
+            # Absolute cycle counts begin at 0
+            return int(math.floor(hc / self.__hc_tick_period))
 
         ## With a given hypercycle, compute a hypercycle that represents the
         #  next local cycle.
         #  @param hc Hypercycle to use for computation
-        #  @return hypercycle representing (local cycle + 1) where local_cycle
+        #  @return hypercycle representing (local cycle) where local_cycle
         #  is computed by the \a hc argument.
         #
-        #  Implemented as LocalToHypercycle(1+HypercycleToLocal(hc))
+        #  Implemented as LocalToHypercycle(HypercycleToLocal(hc))
         def NextLocalCycle(self, hc):
             assert hc is not None
-            return self.LocalToHypercycle(1 + self.HypercycleToLocal(hc))
+            return self.LocalToHypercycle(self.HypercycleToLocal(hc))
 
```
Can you add it?

robbellphd:
Done.

klingaard:
We've fixed the last regression failure with this branch.  We have a couple of major updates to our models tomorrow, so we're planning on moving to this next Tuesday.  My plan:

1. On Tuesday, approve and merge this PR into master of sparta
2. Update our models
3. Tag Sparta with v1.0.0 and make a package -- it's about time I did that and this is a good starting point

Thanks for the patience, Rob!

robbellphd:
Knute, sounds good, no worries!

knute-sifive (approved):

PR #269, opened 2021-08-24
Knutel/issue 268 sparta shared pointer fix

### LINKED ISSUE #268: Cannot create a SpartaSharedPointer instance inside a class that typedefs one to itself
Best to show an example.  You can't do this:
```    
class Foo {
    public:
        using FooPtrType = sparta::SpartaSharedPointer<Foo>;

        void setFoo (const FooPtrType & foo) { foo_ = foo; }

    private:
        FooPtrType foo_;
    };
```
`sizeof` in the nested allocator inside `SpartaSharedPointer` complains of the incomplete type.

knute-sifive:
This addresses the issue that a user cannot instantiate a SpartaSharedPointer inside a class of the same type.  I.e.  `class Foo` (before) could not contain a `SpartaSharedPointer<Foo>` inside it.  

In this PR, I split apart the `SpartaSharedPointerAllocator` from `SpartaSharedPointer`.  So the API *has changed* in this PR.
Old:
```
sparta::SpartaSharedPointer<MyObj>::SpartaSharedPointerAllocator allocator_instance;
```
New:
```
sparta::SpartaSharedPointerAllocator<MyObj> allocator_instance;
```
Everything else is pretty much the same.

closes #268


klingaard (approved):

2henwei:
Awesome!

PR #272, opened 2021-09-03
Argos: Add font size setting for playback bar in View Settings menu

bdutro:
Instead of playing whack-a-mole with different DPIs and font rendering engines, this PR adds a setting that allows users to manually set their desired font size in the Argos UI.


benoy-sifive (approved):
Tested locally
Very helpful on MacOS where the text was too small to read

PR #285, opened 2021-10-15
ReplacementIF extensions for user-defined way order.

robbellphd:
Code changes to allow user-defined way ordering for cache replacement policies and invalid way selection.

### DIFF
--- a/sparta/cache/Cache.hpp
+++ b/sparta/cache/Cache.hpp
@@ -174,6 +174,12 @@ namespace sparta
                 return peekCacheSet(addr).findInvalidWay();
             }
 
+            // Search for invalid in user-defined order of ways.
+            uint32_t findInvalidWay(uint64_t addr, std::vector<uint32_t> way_order) const
+            {
+                return peekCacheSet(addr).findInvalidWay(way_order);
+            }
+
             uint32_t getNumWays() const {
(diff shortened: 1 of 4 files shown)


sparta/cache/Cache.hpp
```diff
@@ -174,6 +174,12 @@ namespace sparta
                 return peekCacheSet(addr).findInvalidWay();
             }
 
+            // Search for invalid in user-defined order of ways.
+            uint32_t findInvalidWay(uint64_t addr, std::vector<uint32_t> way_order) const
```
benoy-sifive, line 178: Make parameter "const std::vector<uint32_t> way_order& way_order)" to avoid the copy

sparta/cache/ReplacementIF.hpp
```diff
@@ -34,18 +34,22 @@ namespace sparta
 
             // Touch way as LRU
             virtual void           touchLRU(uint32_t way) = 0;
+            virtual void           touchLRU(uint32_t way, std::vector<uint32_t> way_order) = 0; // user-defined way order
```
benoy-sifive, line 37: Make parameter "const std::vector<uint32_t> way_order& way_order)" to avoid the copy

Same for functions below...

PR #288, opened 2021-11-01
Bug fix to ValidValue to allow copies of non-valid ValidValues

### LINKED ISSUE #287: ValidValue errors on copy/assignment of another ValidValue
A `ValidValue` shouldn't die if you copy it and it's not valid.  It should copy the validity first, check it, and then make the decision to copy the underlying object.   This comes about when `std::vector`s are used and they resize.

Rules:
1. If the rvalue is a ValidValue type, and it is not valid, only the validity is copied, which is the value false
2. If the rvalue is a ValidValue type, and it is valid, both the validity and the value are copied
3. If the rvalue is a contained type (type of the ValidValue template parameter), it is assumed to be valid, converted to a ValidValue and rule #2 applied

klingaard:
closes #287

### DIFF
--- a/sparta/test/ValidValue/ValidValue_test.cpp
+++ b/sparta/test/ValidValue/ValidValue_test.cpp
@@ -71,6 +71,17 @@ int main()
     sparta::utils::ValidValue<Foo> foo_type2(10, "hello");
     EXPECT_TRUE(foo_type2.isValid());
 
+    sparta::utils::ValidValue<Foo> foo2(15);
+    EXPECT_TRUE(foo2.isValid());
+    foo2.clearValid();
+    EXPECT_FALSE(foo2.isValid());
+    sparta::utils::ValidValue<Foo> foo3 = foo2;
+    EXPECT_FALSE(foo3.isValid());
+
+    std::vector<sparta::utils::ValidValue<Foo>> item;
+    for(uint32_t i = 0; i < 100; ++i) {
+        item.emplace_back(i, "test");
+    }
 
     REPORT_ERROR;
     return ERROR_CODE;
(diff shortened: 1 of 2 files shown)


sparta/test/ValidValue/ValidValue_test.cpp
```diff
@@ -83,6 +83,14 @@ int main()
         item.emplace_back(i, "test");
     }
 
+    std::vector<sparta::utils::ValidValue<Foo>> item2;
+    for(uint32_t i = 0; i < 100; ++i) {
+
```
robbellphd, line 88: A couple of spurious changes got in here...

sparta/test/ValidValue/ValidValue_test.cpp
```diff
@@ -71,6 +71,25 @@ int main()
...
+    EXPECT_TRUE(foo2.isValid());
+    foo2.clearValid();
+    EXPECT_FALSE(foo2.isValid());
+    sparta::utils::ValidValue<Foo> foo3 = foo2;
+    EXPECT_FALSE(foo3.isValid());
+
+    std::vector<sparta::utils::ValidValue<Foo>> item;
+    for(uint32_t i = 0; i < 100; ++i) {
+        item.emplace_back(i, "test");
+    }
```
jackjchen, line 84: Perhaps intended to include?
 `EXPECT_TRUE(item.at(i).isValid());`

robbellphd (commented):
Everything else looks good to me!

sparta/test/ValidValue/ValidValue_test.cpp
knute-sifive: Fixed

sparta/test/ValidValue/ValidValue_test.cpp
knute-sifive: Fixed

knute-sifive (approved):

jackjchen (approved):

robbellphd (approved):

PR #295, opened 2021-12-23
Avoided double reclaiming proxy in PayloadEvent

2henwei:
I ran into the assertion in `reclaimProxy_`.

In my use case, I was canceling DataOutPort calls in the same registered handler.
```
out_port.registerHandler(handlePortCall);

void handlePortCall(T Data)
{
  out_port.cancelIf(criteria);
}
```

I noticed proxy was reclaimed when Scheduler was delivering payload. So it is possible that the proxy can be repeatedly reclaimed if criteria matches. I added an if-condition to skip reclaiming if it was already reclaimed.


knute-sifive:
I'll go ahead and approve this for now, but I think a better approach might be to see if the PLE is currently executing, and if so, do not reclaim it as it will be reclaimed automatically.   The worry I have is that the original assert will no longer catch double free's.

knute-sifive (approved):

PR #297, opened 2021-12-30
DMI redo

### LINKED ISSUE #296: DMI in memory interfacing requires a small refactoring
The DMI class is not conducive to smart invalidation, copying, nor is it really encapsulated.  Needs a revisit, cleanup and actual testing.

knute-sifive:
DMI wasn't really completed nor tested.  This PR cleans up DMI and actually designs/tests it.

closes #296


sparta/sparta/memory/BlockingMemoryIF.hpp
```diff
@@ -9,13 +9,13 @@
...
 
 namespace sparta
 {
     namespace memory
     {
         class TranslationIF;
+        class DMIBlockingMemoryIF;
 
         /*!
          * \brief Pure-virtual memory interface which represents a simple,
```
kathlenehurt-sifive, line 21: This comment needs to be updated since this class isn't pure virtual.

kathlenehurt-sifive (approved):

sparta/sparta/memory/BlockingMemoryIF.hpp
klingaard: Hmmm... the methods `try[Read,Write]` are pure virtual...

sparta/sparta/memory/DMIBlockingMemoryIF.hpp
```diff
@@ -1,84 +1,100 @@
...
 
-    addr_t getAddr() const
-    {
-        return dmi_.getAddr();
-    }
+            //! \return The internal raw pointer.
+            //!
+            //! This is dangerous to use directly as memory bounds
+            //! checking can easily be bypassed
+            void * getRawDataPtr() { return raw_pointer_; }
```
2henwei: I think the assumption made for raw pointer is that the data layout of raw data has to be continuous, though I can't come up with a case which isn't.

sparta/test/Memory/Memory_test.cpp
```diff
@@ -793,3 +795,143 @@ void testMemoryObjectSizes() {
...
+    for (uint32_t i = 0; i < BLOCK_SIZE; ++i) {
+        EXPECT_EQUAL(buf[i], i);
+    }
+
+    // Some silly things -- the following lines of code get around the
+    // private getDMI method in DMIBlockingMemoryIF
+    sparta::memory::BlockingMemoryIF * dmi_blocking_if = dmi;
+    EXPECT_THROW(dmi_blocking_if->getDMI(0, BLOCK_SIZE));
+
+    membif.invalidateAllDMI();
```
2henwei, line 932: Wondering what's the scenario to invalidate DMI ? Does it mean users have to check the availability of DMI before accessing it?

sparta/sparta/memory/DMIBlockingMemoryIF.hpp
klingaard: It's guaranteed to be continuous _only within the BLOCK_SIZE_.  I can update the comments.  If you try to access memory using this pointer beyond the block size, you're gonna find trouble.

sparta/test/Memory/Memory_test.cpp
klingaard: Technically you should always check.   The scenario mostly includes software moving memory around and updating memory maps.  If you originally got a DMI in the core to say ROM @ 0x1000, and software remaps the ROM to 0x3000 and puts something else at 0x1000, the DMI pointer the core has is technically invalid.  It's a DMI pointer to 0x1000, which _still points to ROM_ now at 0x3000.  If you want to access memory at 0x1000, you need to get a new DMI.

sparta/sparta/memory/DMIBlockingMemoryIF.hpp
2henwei: Big trouble!

sparta/test/Memory/Memory_test.cpp
2henwei: Gotcha. We're talking about physical memory right? Didn't have much experience in this, but is it possible to change the memory map of devices in runtime?

klingaard: Yep.  Some memory controllers I've worked on had programmable maps.

PR #300, opened 2022-01-08
Adds ExportedPort class (issue 172)

### LINKED ISSUE #172: Hierarchical block boundary interfacing
Hierarchical block boundary crossing requires additional ports and manual driving. 

For example a super-block "S" has two sub-blocks "SA" and "SB". Another super-block "X" interfaces one data port to S and another credit port back from S to X. 

X->S (data)
S->X(credit)

The super-block just forward the data to say one sub-block "A".  Similarly, return the credit from SA to S to X.

X->S-->SA (data)
SA-->S->X (credit)

S each tick forward the data from X to SA and from SA to X. S interfaces are just used forwarding. 

I think systemC had something like sc_export. I was wondering if there is something like this already exists in SPARTA that I am aware of which helps automatic boundary crossing?

klingaard:
Initial support of an ExportedPort requested in issue #172 

Also added some util functions and fixed some documentation.

closes #172


sparta/sparta/ports/PortSet.hpp
```diff
@@ -47,7 +47,7 @@ namespace sparta
          * \return Pointer to the port or nullptr if not found
          */
         Port * getPort(const std::string & named_port) const {
-            for(uint32_t i = 0; i < Port::N_DIRECTIONS; ++i) {
+            for(uint32_t i = 0; i <Port::N_DIRECTIONS; ++i) {
```
2henwei, line 50: Magic space disappears and it could be `Port::Direction::N_DIRECTIONS`

sparta/sparta/ports/PortSet.hpp
```diff
@@ -108,7 +108,7 @@ namespace sparta
 
     private:
         //! The registered ports within this PortSet
-        RegisteredPortMap registered_ports_[Port::N_DIRECTIONS];
+        RegisteredPortMap registered_ports_[static_cast<uint32_t>(Port::N_DIRECTIONS)];
```
2henwei, line 111: `Port::Direction::N_DIRECTIONS`

sparta/sparta/ports/ExportedPort.hpp
```diff
@@ -0,0 +1,106 @@
...
+
+        //! \brief Override Port::bind
+        //! \param port The port to bind the internal exported port
+        void bind(Port * port) override final;
+
+    private:
+        // The exported port -- to either be found or provided
+        sparta::Port * exported_port_ = nullptr;
+
+        // Non-const as the Port TreeNode contained in the path will be modified
```
2henwei, line 120: I'm wondering this ... could you elaborate it more?

2henwei (approved):
This looks good to me. I was concerning that ExportedPort doesn't implement public methods of other derived classes (e.g. `cancel()` in DataPort), but eventually realized that ExportedPort acts like a proxy to help users bind underlying ports. I think we should add some comments like this: `ExportedPort` is only used for binding in high-level Sparta Units, not for normal usage of Sparta Port.

2henwei:
Should we invalidate other public methods of Port (e.g. isBound(), getDirection()) in ExportedPort?

sparta/sparta/ports/PortSet.hpp
klingaard: Fixed.

sparta/sparta/ports/PortSet.hpp
klingaard: Fixed.

klingaard:
> This looks good to me. I was concerning that ExportedPort doesn't implement public methods of other derived classes (e.g. `cancel()` in DataPort), but eventually realized that ExportedPort acts like a proxy to help users bind underlying ports. I think we should add some comments like this: `ExportedPort` is only used for binding in high-level Sparta Units, not for normal usage of Sparta Port.

Agreed, but I should add more testing than what I have.  The public APIs of Port should be available to use as well.

klingaard:
Hmmm... this gets tricky. I can promote this class as a proxy (full service of a Port) or just make it a facade (limited capabilities).  Tough one.  I think I'll go for the latter (facade) and add a method to get the internal port if a developer wants to use the full capabilities of a port.

sparta/sparta/ports/ExportedPort.hpp
klingaard: I updated the comments:
```
        // Non-const as the Port TreeNode contained in the path will
        // be modified during binding.  If this variable were const,
        // the code could not find the to-be-modified internal_port_
```

klingaard:
> Should we invalidate other public methods of Port (e.g. isBound(), getDirection()) in ExportedPort?

I'm thinking no.  This port class is a simple wrapper around an internal port.  I added methods to get to the internal port to make those queries, but the framework in general doesn't require it.

2henwei:
I'd also override those methods
```
         Direction getDirection() const {
             if(false == isBound()) {
                 return Port::Direction::UNKNOWN;
             }
             return internal_port_->getDirection();
         }
  
         void participateInAutoPrecedence(bool participate) {
             dabble_assert(isBound(), "You haven't bound this ExportedPort ...");
             internal_port_->participateInAutoPrecedence(particpate);
         }
  
         bool doesParticipateInAutoPrecedence() const {
             if(false == isBound()) {
                 return false;
             }
             return interal_port_->doesParticipateInAutoPrecedence();
         }
```

knute-sifive:
I'd agree with 
```         
Direction getDirection() const;
```
But not the other two.  The ExportedPort should not have a say in how the precedence rules apply to a port deep into the simulator.

knute-sifive:
Ok, fine, you win.  I learned I cannot `dynamic_cast` a Port to an ExportedPort inside an ExportedPort method 'cause of my private inheritance.  Drat.  I'll make these functions virtual and provide error messages.

Issue #145, opened 2020-05-22
unable to create environment from rendered conda recipe on Ubuntu


MichaelSchoenfelder-SiV:
When executing `./scripts/create_conda_env.sh sparta dev` on Ubuntu 16, conda complains that it doesn't know how to merge X and Y for a variety of packages where the last `=id` differs between the packages.  If I remove all the conflicts by deleting one of the duplicates from the environment file, the script finishes.  However, the environment can't be created due to multiple unresolvable conflicts.  Only after removing the build id pins was I able to create the environment.

@timsnyder

klingaard:
So, Mike, do you have a solution to push on this?

MichaelSchoenfelder-SiV:
No.  @timsnyder is familiar with the problem and had already built some workarounds into the flow.  I am not sure why they didn't work.  It may be a problem in conda itself.  My solution at SiFive was to manually manipulate the recipe until it worked, but that isn't faithful to the intent of the solution.

MichaelSchoenfelder-SiV:
Other conda maintenance: We also need to add the lld package to support LTO.  And it looks like we might be able to bump the LLVM to v11 since v11 is now available in conda forge.  I will be testing LLVM in a semi-manually-built conda environment.

jackjchen:
@MichaelSchoenfelder-SiV , I ran into a similar issue on OSX 11.5.2. Changed some versions to get a rendered recipe, but the solver had numerous conflicts. Do you have a working environment yaml to share?

PR #310, opened 2022-02-24
Add CMake build types and fix version requirements

jackjchen:
- Add support for the other cmake standard build types: RelWithDebInfo and MinSizeRel
> see https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html
> This enables a clean choice of setting optimization level of -O2 by building RelWithDebInfo

- Update cmake files with required versions of cmake to be consistent as documented in #301 

- Manually add Travis CI VM size to try to reduce time required and help with timeout issue


benoy-sifive:
Could the cmake_minimum_required (VERSION 3.15) piece live at the top level somewhere instead of in many subdirs?

klingaard:
Thanks Jack for helping with the timeout issue.

klingaard (approved):

klingaard:
> Could the cmake_minimum_required (VERSION 3.15) piece live at the top level somewhere instead of in many subdirs?

That would be nice, if possible

jackjchen:
> Thanks Jack for helping with the timeout issue.
Sure thing. The larger VM has 4 CPUs and 16G (double) which ran faster (charges double the credits too). Unfortunately I think it might be Travis docker container that still sets $CPU_COUNT to 2 (so make -j is still limited). This will take more investigation, but the change somewhat addresses the timeout for now.

jackjchen:
> > Could the cmake_minimum_required (VERSION 3.15) piece live at the top level somewhere instead of in many subdirs?
> 
> That would be nice, if possible

I did identify a potential solution. It would be to create a "map/cmake/map-config.cmake" file that sets a common cmake min version. Then the 6 CMakeLists.txt files would each "include" the config file in the first line. Minor hesitancy to introduce an upwards top-level file dependency for the currently self-contained subdirs.  It may be worthwhile when there is more than the single config value to share and more so if there is config anticipated to change often.

timsnyder-siv:
Sorry I missed your question Jack. I don't remember why we even have anything running on Travis. I vaguely remember hearing that they slashed their time limits for open source jobs and I think it might be good to move to something else.

Should just have to change the conda-forge.yaml and run `conda smithy rerender` to update the ci configuration files. Be aware that there might be more than you need. I think there are new files that aren't in the skip render list. I'm happy to help with this.

timsnyder-siv:
I looked back and found https://github.com/sparcians/map/commit/f4d2e29439eeafabd5b5b364fe2dcad713e03d05 which shows me moving the linux builds from azure to travis when we got started.  I said that we needed more diskspace at that time.  I'm not sure why we were running into the diskspace being an issue on those runners.  I wonder if it would still be an issue.

jackjchen:
Thanks Tim. Any suggested CI solution for linux would be very helpful. Travis does seem questionable long term. Previously looked at  yaml and rerender changes from MacOS but still had trouble creating a conda env from the rendered recipe.

timsnyder-siv:
>  but still had trouble creating a conda env from the rendered recipe.

I'm curious to know more about what this means.  I think you've told me in the past but I forget.  

> Any suggested CI solution for linux would be very helpful.

I'd probably start with moving back to Azure.  I also see support in conda-smithy for github-actions now.  Could maybe try that.

knute-sifive:
Have you looked at CircleCI?

jackjchen:
> > but still had trouble creating a conda env from the rendered recipe.
> 
> I'm curious to know more about what this means. I think you've told me in the past but I forget.

Previous discussion is in #145 
I'll have to revisit to find the current behavior and details.
Background, is it looks like I was testing `travis.yml` with removing version pinning for numpy and clang (I know we probably need pin to 10?), and removing python 3.7. But then the `create_conda_env.sh` from the command line seems to have problems with the rendered recipe (maybe even after removing variant hash/ids).

jackjchen:
> Have you looked at CircleCI?

I have not. Tim, if you have a conda-smithy supported solution, I can help test.

timsnyder-siv:
I haven't looked at circleci but CircleCi is one of the supported providers.  Instead of travis at https://github.com/sparcians/map/blob/5c3ac01301664a4794e38faaa1d0bae07d3f16e0/conda-forge.yml#L17 put `circle` (see https://conda-forge.org/docs/maintainer/conda_forge_yml.html#provider but note that `github-actions` is undocumented still for some reason)

Then, you have to rerender.

Finally, you have to figure out how to setup a 'sparcians' account at circle and hook it into the github repo (for it to be able to call back and update PRs with checks results).

And when you open a PR with your new stuff, it should hopefully run on circle.  You'll have to refer to circleci docs on how to do that.  Let me know if you get stuck.

jackjchen:
Thanks Tim, I'll try it out and open a PR. If no one is blocked on current status, I will have more time to work on this in a couple of weeks.

klingaard:
With my latest C++ updates, the MacOS is building and running again -- success!  Now onto Linux builds...  

Looking forward to your PR, Jack!

PR #312, opened 2022-03-18
Knutel/doc cleanup

klingaard:


klingaard:
Ugh... it looks like the latest clang finally did away with the `std::iterator` boiler.  Guess I need to modify Sparta to explicitly define these type traits.  Grumble.

knute-sifive (approved):

PR #315, opened 2022-03-24
Put in a fix to not dump reports when the sim crashes

knute-sifive:
Also, this honors the `--report-on-error` flag


knute-sifive:
closes #315

klingaard (approved):

sparta/src/Simulation.cpp
```diff
@@ -933,6 +933,7 @@ void Simulation::run(uint64_t run_time)
 
         // Rethrow exception if necessary
         if(eptr != std::exception_ptr()){
```
benoy-sifive, line 935: eptr could use a better name

benoy-sifive (approved):

PR #317, opened 2022-03-27
Bump compilers and move linux to circle

timsnyder:
I made manual changes in the first three commits to `conda-forge.yml` and `conda.recipe/conda_build_config.yaml` and `conda.recipe/meta.yaml` to:
* run CI with two versions of the compilers.  Absolute latest and the version that conda-forge is currently using to build everything.  I just wanted to demonstrate again how we can do multiple compiler versions.  I'm hopeful that we will be able to build with the latest and greatest and can possibly remove the pinnings in map and just float along with conda-forge
* change linux CI to circle.  I did some clicky clicky stuff in the web UI to get circle hooked up to sparcians org and to pay attention to the map repo.  Since we already had a `.circleci/config.yml` (that was populated by conda-smithy before to be empty) it saw when I pushed a new branch and tried to run the new workflows.
* I noticed that the version of Boost also bumps to 1.74 with this change because conda-forge has moved forward.

Then, I ran `conda smithy rerender` (you have to have `conda-smithy` installed in your env and make sure it is updated to have the latest version of it as well as `conda-forge-pinnings`) to generate the changes in the .ci files.


timsnyder:
This obviously isn't ready to merge because circle is failing to even run the docker container.  I also noticed that it is configured to try to upload the final conda package to anaconda.org and we don't want that to happen.  I have changes that I'll push but I'm curious to see what the osx build on azure does first.

timsnyder:
Not sure where I saw clang 14 yesterday but it appears to have been removed or I was looking at one of the prereq packages.  It's only been released for like 4 days at https://github.com/llvm/llvm-project/releases/tag/llvmorg-14.0.0 so I'm not surprised.  I'll bump my concept of "absolute latest available" back to 13.  This is why the bleeding edge osx build failed.

timsnyder:
https://app.circleci.com/pipelines/github/sparcians/map?invite=true is where you go to see the builds on circleci.  Let me know once you've setup your accounts over there and I can grow the list of people who have "admin" privs or whatever we need to do.

klingaard:
Hey Tim, this is AWESOME!  Thanks for putting the time into this.  I've set up my account on CircleCI and started following map.

klingaard:
Do you need me to look at the MacOS fails?  Seems one regression test is barfing...

timsnyder:
I was going to add `CTEST_OUTPUT_ON_FAILURE=1` and rerun to see what the deal is.  I got a bit sidetracked trying to PR some fixes to the 'conda-smithy' script that generates the CI scripts.  Let me push the thing that will print the failing test output to the CI.

timsnyder:
I was also thinking that since it seems to be passing with the clang13, I don't know if you want to regress multiple versions of the compilers all the time or if we should just run CI usually on whatever the defaults are from conda-forge and from time to time we can run a workflow that checks to see if there are new versions of things and runs a CI with those?

timsnyder:
Oh, and I definitely didn't mean to have linux running four builds.  I'm burning through the circleci free credits pretty fast by doing that.

timsnyder:
Seems like that Failing test was just flaky.  I'm going to push some commits to prune down the matrix to just one build for mac and one for linux and then this should be gtg.

klingaard:
I like the idea of running multiple versions of clang to find the next C++ issue that's been hidden by older compilers -- and if that means using the defaults that come with condo-forge, I'm good with that.  

I assume this only runs with a PR and not with every branch?  

Again, thanks a million for putting this together.

klingaard (approved):

timsnyder:
I think right now it runs with every branch.  I can look into fixing that.

klingaard:
PR to master branch is only what I want.  Post merge isn’t necessary.

Thanks!

timsnyder:
Oooohhhh Kay.  I think I have this figured out.

This required some config that is outside of the sparcians/map repo.  That's both good and bad.  Good because `conda smithy rerender` won't be able to clobber it.  Bad because it will be hard to remember where the heck we went to configure this stuff.  So, I'm making some notes here.

Azure
On https://dev.azure.com/sparcians/map/_apps/hub/ms.vss-ciworkflow.build-ci-hub?_a=edit-build-definition&id=1&view=Tab_Triggers I disabled "Continuous Integration" because that's what does the builds on commits to branches.  I left "Pull Request Validation" enabled but overrode the YAML config to say only build PRs to master.

Circle
You have to go into https://app.circleci.com/settings/project/github/sparcians/map/advanced and turn on "only build PRs" but that will still build commits to master.  So, I added the filter described in the blog post https://discuss.circleci.com/t/only-build-pull-requests-not-every-branch/200/8.  `conda smithy rerender` will be able to clobber the changes I made by hand.  I looked into adjusting `conda-smithy` so that it would be able to take the filter but that seems like more work than I want to do right now.

Except I seem to have managed to make circle start failing now.

PR #318, opened 2022-03-28
Update README.md

knute-sifive:


README.md
```diff
@@ -17,6 +17,6 @@ MAP is broken into two parts:
 
 ## Current build status
 
-[![Linux Build Status](https://travis-ci.com/sparcians/map.svg?branch=master)](https://travis-ci.com/sparcians/map/branches)
+[![CircleCI](https://circleci.com/gh/sparcians/map/tree/master.svg?style=svg)](https://circleci.com/gh/sparcians/map/tree/master)
```
timsnyder, line 20: I think since you're not wanting CI to run on merged master ever, you might want to have a slightly different URL for the status badge.  Just a second.  I'll look for what you might want and suggest it.

timsnyder: Yeah, if you look at the "rich diff" of your change, it renders as "failed" "NO BUILDS" for Circle because we've never built master.
image

And the azure badge is a bit misleading because it will always show the last build from master whether it passed or failed.

timsnyder: Also, I think that circleci is using it's 'large' instance.  Seems like it is finishing in half the time as Azure.  I don't know if it would save any credits if you drop it down a size or two.  Again, doesn't seem like you should worry about that until you hit the credit limit.  Speaking of which, can you see https://app.circleci.com/settings/plan/github/sparcians/overview?  I don't see how to add you as an admin of the sparcians org I created.  I'm seriously wondering if it is because they are trying to get us to pay for a plan.  On that page, I see something about 30,000 credits per month but then on another page, I see that if you are a free and open-source project, you get 400k per month.  I need to dig into the docs for that.

knute-sifive: Thanks for looking at that.  Yeah, I see the rich view as borked.   I took your suggestion (cool button).

I don't mind (and would actually prefer) the badging to reflect the state of master and not the latest PR.  Yeah, that means slurping up more credits, but maybe we won't have too many PRs.  😄 

I can see the overview page.  I see that we've used 5K credits so far.  Do we get 30K/month to use?  I hope it resets.

timsnyder:
I suppose that you can always manually build master from time to time if that's what you want to do as well.  Just let me know if you want help changing if/howoften master runs on either of the platforms.

timsnyder (approved):

PR #323, opened 2022-03-31
Update create env

timsnyder:
the environment creation scripts need to be tweaked following #317.  I forgot about them...


knute-sifive (approved):
I noticed the README.md banner still isn't right.   No need to put it in this PR, but if you want to play with it...

PR #327, opened 2022-04-14
Update README.md

benoy-sifive:
Add additional package requirements for Python
Update cmake example for some MacOS envs

### DIFF
--- a/sparta/README.md
+++ b/sparta/README.md
@@ -85,7 +85,6 @@ Open a MacOS Terminal
 
 Additionally, there are Python packages to install
 1. `pip install cython`
-2. `pip install `
 
 Clone sparta through ssh: `git clone ssh://github.com/sparcians/map.git`
 
--- a/sparta/README.md
+++ b/sparta/README.md
@@ -83,6 +83,10 @@ Open a MacOS Terminal
 1. `sudo easy_install Pygments==2.5.2` # Do not use HomeBrew! Needed for cppcheck-htmlreport
 1. Install XCode and check for clang: `clang --version`
 
+Additionally, there are Python packages to install
+1. `pip install cython`
+2. `pip install `
+
 Clone sparta through ssh: `git clone ssh://github.com/sparcians/map.git`
 
 Attempt a build:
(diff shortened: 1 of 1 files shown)


sparta/README.md
```diff
@@ -83,14 +83,19 @@ Open a MacOS Terminal
 1. `sudo easy_install Pygments==2.5.2` # Do not use HomeBrew! Needed for cppcheck-htmlreport
 1. Install XCode and check for clang: `clang --version`
 
+Additionally, there are Python packages to install
+1. `pip install cython`
+2. `pip install `
```
timsnyder, line 88: Why are we using easy_install and pip instead of conda?  Just curious.

sparta/README.md
```diff
@@ -83,14 +83,19 @@ Open a MacOS Terminal
...
 Clone sparta through ssh: `git clone ssh://github.com/sparcians/map.git`
 
 Attempt a build:
 
 1. `mkdir release` (directory name not important)
 1. `cd release`
 1. `CC=clang CXX=clang++ cmake .. -DCMAKE_BUILD_TYPE=Release`
-1. `make`
+2. NOTE: Some MacOS environments will need `cmake .. -DCMAKE_CXX_FLAGS=-isystem\ /usr/local/include -DCMAKE_BUILD_TYPE=Release`
+3. `make`
```
timsnyder, line 98: Why is this only some MacOS environments?  How do you know if you need this?

sparta/README.md
timsnyder: I guess this is for the homebrew setup that people use.  I see.

sparta/README.md
benoy-sifive: I am not sure how I got into this state after updating packages (via brew)
Of course, its only me that hit this currently ... do you think we should remove it or leave it for the obscure cases?

klingaard: For some reason, a `cmake` build with an older MacOS SDK will not include `/usr/local/include` in the system include path.  Haven't researched this, but this fix works for now.

klingaard (approved):

PR #335, opened 2022-05-24
ENH: added CMake option to enable helios/sparta

tvandera:
This allows to disable helios with `cmake -DENABLE_HELIOS=OFF`


knute-sifive (approved):
I like this change.  Mike, we should take advantage for our internal builds.

Thanks!

@MichaelSchoenfelder-SiV

PR #336, opened 2022-05-24
FIX: run git describe in source directory, not build

tvandera:
This fixes the error 
```
fatal: not a git repository (or any of the parent directories): .git
CMake Error at sparta/CMakeLists.txt:109 (message):
  could not run git command 'git describe --tags --always', rc=128
```
when not running `cmake` in the source directory.


knute-sifive (approved):
Good catch.  Much appreciated.

PR #340, opened 2022-06-02
Fixed a bug in overwriting Array element

aka2en:


sparta/sparta/resources/Array.hpp
```diff
@@ -874,6 +874,15 @@ namespace sparta
             sparta_assert(idx < num_entries_,
                         "Cannot write to an index outside the bounds of the array.");
 
+            // We're overwriting an entry, clear valid and age information.
+            if(SPARTA_EXPECT_FALSE(isValid(idx)))
+            {
+                --num_valid_;
+                if(ArrayT == ArrayType::AGED) {
```
klingaard, line 881: Probably could get away with an 
'''
if constexpr 
'''
here, but that will be inconsistent with the rest of the code. (Unless you change them all 😜)

PR #345, opened 2022-06-30
Add conda instructions to sparta/README.md

timsnyder:


jackjchen:
@timsnyder does the above file/simdb issue look like just a spurious bug in osx/azure CI?

knute-sifive:
> @timsnyder does the CI fail around file/simdb look like just a spurious bug in osx/azure?

I think so.  The re-run will be telling...

timsnyder:
I'm curious.  It will be interesting to see what happens on the rerun.  I did notice that `addr2line` is missing from the conda environment, causing the backtrace to look wonky.  Probably should add that as well but I'll open another PR for it.  I'm currently trying to see if conda-forge has gnu binutils available for osx-64 or not.

Knute, an alternative might be to use [atos](https://stackoverflow.com/questions/54320054/how-to-use-atos-or-addr2line-on-mac-os-x) on mac

knute-sifive:
I would _LOVE_ to get the backtrace working again on Mac (and demangling).  I disabled it for now and never got back to it.

knute-sifive (approved):

timsnyder:
> I disabled it for now and never got back to it.
However you 'disabled' it didn't seem to get all of the places where it's called https://dev.azure.com/sparcians/map/_build/results?buildId=846&view=logs&j=58ac6aab-c4bc-5de2-2894-98e408cc8ec9&t=32c9ac78-f8b1-54ea-01a1-c8d72a874093&l=2684

I'll look for addr2line a bit more from conda .  Are you calling `c++filt` pipe stuff through also?  Where's the demangling?  GH doesn't like searching for `c++filt` if it is in there somewhere.

knute-sifive:
> I'll look for addr2line a bit more from conda . Are you calling `c++filt` pipe stuff through also? Where's the demangling? GH doesn't like searching for `c++filt` if it is in there somewhere.

I honestly don't remember how it works.  Maybe @nhaw might have a braincell holding on to history, but it's implemented here: https://github.com/sparcians/map/blob/master/sparta/src/Backtrace.cpp

knute-sifive:
I think Nick originally called some demangline C++ routines.

MichaelSchoenfelder-SiV (commented):
I just realized there is a `.scripts/create_conda*` and a `./scripts/create_conda*`.  My brain has only been seeing one of those.  That explains some of my confusion.

The readme doesn't mention the "run" environment.  I assume the "dev" is a superset and will be sufficient.  People may find a "run" environment a bit more efficient if deploying the environment at runtime.

timsnyder:
> The readme doesn't mention the "run" environment.

Sure.  The readme is about getting developers setup to build sparta.  If you only need to run a built sparta simulator, you can use the run environment.

> `.scripts` in the toplevel 

would be nice if it could be named `.ci_scripts`.  It is generated by `conda smithy rerender` when you ask it to generate the CI infrastructure from `conda_forge.yml` and `conda.recipe` as described in https://conda-forge.org/docs/user/ci-skeleton.html.  It would be interesting to see if a `.scripts/README` would remain unmodified by `conda smithy rerender` so that we could tell people `THESE AREN'T THE SCRIPTS YOU ARE LOOKING FOR` :jedi handwave: 

Some, (but not all) of the generated scripts have nice warnings at the top of them stating that they are generated files.  I'm sure that could be expanded with a pointer to wtf conda-smithy is at conda-forge/conda-smithy

MichaelSchoenfelder-SiV:
Thanks for the explanation.  The new README was what helped me figure out that .scripts was different then ./scripts.  That resolved my confusion when I first tried to help Kevin and looked in .scripts.  Therefore mission accomplished.  Thank you.

Issue #356, opened 2022-08-30
Confusing comments regarding state of SyncPort readiness (SyncPort::isReady())


alexanderFell:
In https://github.com/sparcians/map/blob/afacf1299e451578a7e242c77461995e6af1f02a/sparta/sparta/ports/SyncPort.hpp#L175 the method `isReady` allows specifying an offset from the current clock cycle to check the readiness of the port in the future and the comment for this parameter states: "Number of cycles the sender would use in a send() call when sending." According to the comment, this would be valid:

```
if(syncOutPort->isReady(10)) {
   syncOutPort->send(dataPacket, 10);
}
```

Since we do not know, what the future brings (in fact the bounded incoming port could have `setReady(false)` in the meantime). In case, the port is not ready by the time of sending (here in 10 clock cycles), the packet is recirculated.

However, `isReady` calls `_couldAccept()`.
The problem arises, that _couldAccept asserts that the parameter of `isReady` is 0 in https://github.com/sparcians/map/blob/afacf1299e451578a7e242c77461995e6af1f02a/sparta/sparta/ports/SyncPort.hpp#L768

This contradicts the comment regarding `isReady` and `couldAccept_` in https://github.com/sparcians/map/blob/master/sparta/sparta/ports/SyncPort.hpp#L34-L36: "When trying to send new data, SyncOutPort will then call SyncInPort::couldAccept_() to determine if data could be delivered on the given cycle."

As far as I can see, only `isReady` contains calls to `couldAccept_` but not the `send` method.

I must have misunderstood something. As of now, the code and the comments are contradictory and would appreciate a clarification.

knute-sifive:
I see your quandary and there's definitely a possible contradiction between the default param value and the concept of next ready.  Need to chew on this one for a day and look at the SyncPort test.

knute-sifive:
Ok, I've re-familiarized myself with this code (I wasn't the original author) and I think the SyncPort API was written to address two types of use cases (IIRC): 

1. Fire and forget allowing the SyncPort to send data on alignment between the clocks (via `forwardData_`)
2. A query API to allow the user to figure out when the data can be sent

I agree that the `isReady`'s default value is ... pointless.   

If you want to change/fix/remove the default value (and the docs), I'd be happy to take the PR.  Thanks for pointing it out.

alexanderFell:
[...]

PR #360, opened 2022-09-02
Removing parameter from isReady

alexanderFell:
The parameter for isReady to check the readiness of a connected SyncInPort can be omitted. In the documentation it is suggested, to check for  SyncOutPort::isReady(n), if this is followed by a SyncOutPort::send(message, n). However, between the current clock cycle `c` and `c+n`, the input port can become unready by calling SyncInPort::setReady(false). Therefore, there is a discrepancy of what the method `isReady` does and the comment. 
Further, `isReady` calls `couldAccept_` which then asserts that the parameter given to `isReady`, equals 0. 
This commit removes the parameter from the list of `isReady` and `couldAccept_`. Further, it improve comments regarding the purpose and functionality of `isReady`.

Related to issue #356

### DIFF
--- a/sparta/sparta/ports/SyncPort.hpp
+++ b/sparta/sparta/ports/SyncPort.hpp
@@ -3,7 +3,7 @@
 /**
  * \file   SyncPort.hpp
  *
- * \brief  File that defines syncrhonized input/output ports.
+ * \brief  File that defines synchronized input/output ports.
  *
  * Explanation of ready/valid mechanism:
  *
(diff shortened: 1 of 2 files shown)


alexanderFell:
For consideration, this commit changes the parameters of `SyncInPort::isReady()`. Hence, maybe instead of removing it, the API should be declared deprecated first to prevent breaking current projects.

knute-sifive:
Yeah, this make a lot more sense.  First of all, general query methods _should not assert_ on the user with their own API.  Secondly, the comments are very good/informative.  Much appreciated.

knute-sifive (approved):

PR #362, opened 2022-09-07
Fix for allow_slide setting for SyncPorts, in case of a collision

alexanderFell:
This commit resolves 2 issues with the `allow_slide` parameter in `SyncOutPort::send` and derivatives (`SyncOutPort::sendAndAllowSlide()`):

1. The documentation does not clearly explain the impact and function of `allow_slide`
2. Setting `allow_slide` to true has either no effect or in case of a collision causes a `sparta_assert`, which should not happen.

### Documentation
The improvement in documentation explains what the parameter `allow_slide` does. This is related to issue #356 in which it has been briefly mentioned.

### `allow_slide`
The SyncPort allows the transmission of 1 packet per clock cycle only. Hence, calling `send` for the same SyncOutPort twice within a clock cycle, will cause an assertion to be thrown. The `allow_slide` = true allows to reschedule packets which are sent in the same clock cycle. The newly added packets will recirculate and rescheduled for transmission in the next possible clock cycle after all the packets in flight have been transmitted. Consider the following example code:
```
if(outPorts->isReady()) {
    outPorts->send(message0, 1, false);
    outPorts->send(message1, 1, false);
    outPorts->send(message2, 1, false);
}
```
will result in
```
21: top.arch.memory_cpu0.crossbar.ports.out_arbiter0: SEND @22 allow_slide=0 # 0x564c98ed2070
21: top.arch.memory_cpu0.ports.in_crossbar_to_llc: RECEIVE SCHEDULED @22(0)  # 0x564c98ed2070
21: top.arch.memory_cpu0.crossbar.ports.out_arbiter0: SEND @22 allow_slide=0 # 0x564c98ed2071
21: top.arch.memory_cpu0.ports.in_crossbar_to_llc: RECEIVE SCHEDULED @22(1)  # 0x564c98ed2071
terminate called after throwing an instance of 'sparta::SpartaException'
  what():  prev_data_arrival_tick_ < abs_scheduled_tick || prev_data_arrival_tick_ == PREV_DATA_ARRIVAL_TICK_INIT: top.arch.memory_cpu0.ports.in_crossbar_to_llc: attempt to schedule send for tick 34, which is not later than the previous data at tick 34; SyncInPorts should only get data once per cycle; data was: 0x564c98ed2070: in file: '../map/sparta/sparta/ports/SyncPort.hpp', on line: 944
```
The first 2 lines schedule the first `send` call correctly. The second `send` call (for data with the address `0x564c98ed2071`) cause the assertion.

The problem arises, that in the code snipped above, setting the last parameter in `send` to `true` to activate the `allow_slide` mechanism, the output is the same, which is incorrect. Instead, the `message1` and `message2` in the second and third call to `send` should be scheduled for delivery in clock cycle 23 and 24, respectively, following `message0` in clock cycle 22. With this commit, the behavior is correct:

```
21: top.arch.memory_cpu0.crossbar.ports.out_arbiter0: SEND @22 allow_slide=1 # 0x556d6b022070
21: top.arch.memory_cpu0.ports.in_crossbar_to_llc: RECEIVE SCHEDULED @22(0)  # 0x556d6b022070
21: top.arch.memory_cpu0.crossbar.ports.out_arbiter0: SEND @23 allow_slide=1 # 0x556d6b022071
21: top.arch.memory_cpu0.ports.in_crossbar_to_llc: RECEIVE SCHEDULED @23(1)  # 0x556d6b022071
21: top.arch.memory_cpu0.crossbar.ports.out_arbiter0: SEND @24 allow_slide=1 # 0x556d6b022072
21: top.arch.memory_cpu0.ports.in_crossbar_to_llc: RECEIVE SCHEDULED @24(2)  # 0x556d6b022072
22: DELIVERING @22(2)  # 0x556d6b022070
23: DELIVERING @23(1)  # 0x556d6b022071
24: DELIVERING @24(0)  # 0x556d6b022072
```

The commit fixes the rescheduling by checking, how many packets are already in flight and adding this number to the desired delay given in `send`

Points worth noticing/discussing and Terms and Conditions:
1. Currently, there is no limit on packets that are in flight. Hence, the implementation cannot be used to simulate FIFO buffers having upper bounds, which would be more realistic in hardware. To have FIFOs for the packets in flight, a bit of infrastructure surrounding the SyncPort pair needs to be implemented.
2. I have not tested cross clock domain behavior of the slide, as my current project has only 1 common clock.
3. It is unclear to me (maybe I am overthinking it), how the SyncPort should react in case the `send` call is interleaved with various differentiating delay parameters. For instance, imagine that the second `send` in the snippet above is replaced by `outPorts->send(message1, 1000, true);` I believe that the sending of packets needs to be always in ascending order. Hence, `message2` can only be sent at a delay of > 1000.

### DIFF
--- a/sparta/sparta/ports/SyncPort.hpp
+++ b/sparta/sparta/ports/SyncPort.hpp
@@ -331,7 +331,7 @@ namespace sparta
             // Convert the absolute tick of the send event into a current
             // cycle relative cycle and return that value
             Clock::Cycle next_send_cycle = clk_->getCycle(send_tick) + send_delay_cycles;
-            sparta_assert(next_send_cycle > current_cycle);
+            sparta_assert(next_send_cycle >= current_cycle);
             return next_send_cycle;
         }
 
@@ -855,7 +855,7 @@ namespace sparta
                                                    send_clk->getTick(send_delay_cycles),
                                                    send_clk, receive_delay_ticks_, receiver_clock_);
 
-            sparta_assert( data_arrival_tick > num_delay_ticks );
+            sparta_assert( data_arrival_tick >= num_delay_ticks );
             return num_delay_ticks;
         }
 
--- a/sparta/sparta/ports/SyncPort.hpp
+++ b/sparta/sparta/ports/SyncPort.hpp
@@ -229,7 +229,10 @@ namespace sparta
         }
 
         /**
-         *  \brief Send data on the output port and allow slide
+         *  \brief Send data on the output port. In case of a conflict (multiple packets scheduled
+         *  to be received within the same clock cycle), the new packet is rescheduled for transmission
+         *  in the next clock cycle automatically. There is no limit on how many packets can be
+         *  rescheduled, the FIFO is without an upper bound.
          *  \param dat The data to send
          *  \param send_delay_cycles Cycles to delay before sending
          *
@@ -244,8 +247,12 @@ namespace sparta
          *  \brief Send data on the output port
          *  \param dat The data to send
          *  \param send_delay_cycles Cycles to delay before sending
-         *  \param allow_slide Allows the receive to slide relative to
-         *                     previous requests
+         *  \param allow_slide If set to true and, in case of packets already in flight and scheduled to be
+         *                     delivered in the same clock cycle, automatically reschedule this new packet to be
+         *                     transmitted in the next following clock cycle. If set to false, no rescheduling is
+         *                     performed. In case of of a collision, the port will throw an assertion.
+         *                     There is no limit on the number of packets that can be rescheduled as the FIFO
+         *                     does not have an upper bound.
          *
          *  \return The delay in ticks from sending
          */
--- a/sparta/sparta/ports/SyncPort.hpp
+++ b/sparta/sparta/ports/SyncPort.hpp
@@ -93,7 +93,7 @@ namespace sparta
      *  a bus.  The destination SyncInPort will handle any latch
      *  delays.  Note that since this potentially send occurs across a
      *  clock boundary, the number of cycles actually delayed won't
-     *  necessarily be in the sending clock comain.
+     *  necessarily be in the sending clock domain.
      *
      *  The rules for sending across clock boundaries are:
      *
@@ -261,11 +261,7 @@ namespace sparta
             sparta_assert(sync_in_port_ != 0, "Attempting to send data on unbound port:" << getLocation());
             sparta_assert(clk_->isPosedge(), "Posedge check failed in port:" << getLocation());
 
-            Clock::Cycle send_cycle = clk_->currentCycle() + send_delay_cycles;
-            
-            if(allow_slide) {
-                send_cycle += sync_in_port_->getNumInFlight();
-            }
+            Clock::Cycle send_cycle = computeNextAvailableCycleForSend(send_delay_cycles, 0);
             
             if (SPARTA_EXPECT_FALSE(info_logger_)) {
                 info_logger_ << "SEND @" << send_cycle
@@ -334,11 +330,9 @@ namespace sparta
 
             // Convert the absolute tick of the send event into a current
             // cycle relative cycle and return that value
-            Clock::Cycle next_send_cycle = clk_->getCycle(send_tick);
+            Clock::Cycle next_send_cycle = clk_->getCycle(send_tick) + send_delay_cycles;
             sparta_assert(next_send_cycle > current_cycle);
-            Clock::Cycle delay_cycle = next_send_cycle - current_cycle;
-
-            return delay_cycle;
+            return next_send_cycle;
         }
 
 
--- a/sparta/sparta/ports/SyncPort.hpp
+++ b/sparta/sparta/ports/SyncPort.hpp
@@ -255,6 +255,11 @@ namespace sparta
             sparta_assert(clk_->isPosedge(), "Posedge check failed in port:" << getLocation());
 
             Clock::Cycle send_cycle = clk_->currentCycle() + send_delay_cycles;
+            
+            if(allow_slide) {
+                send_cycle += sync_in_port_->getNumInFlight();
+            }
+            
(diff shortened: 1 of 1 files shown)


sparta/sparta/ports/SyncPort.hpp
```diff
@@ -255,6 +262,11 @@ namespace sparta
             sparta_assert(clk_->isPosedge(), "Posedge check failed in port:" << getLocation());
 
             Clock::Cycle send_cycle = clk_->currentCycle() + send_delay_cycles;
+            
+            if(allow_slide) {
+                send_cycle += sync_in_port_->getNumInFlight();
+            }
```
klingaard, line 268: Wonder if this code is best placed in the inport's `send_` method or better yet `computeSendToReceiveTickDelay_`.  Thoughts?

knute-sifive: Ok, let me know how you want to proceed with this.

alexanderFell: I had a look at `computeNextAvailableCycleForSend` which should compute the next available clock cycle for a packet to be sent. However, I am observing a rather curious behavior, when it is used:

```
21: top.arch.memory_cpu0.crossbar.ports.out_arbiter0: SEND @22 allow_slide=1 # 0x5565f39b6070, 
21: top.arch.memory_cpu0.ports.in_crossbar_to_llc: RECEIVE SCHEDULED @22(0)  # 0x5565f39b6070

21: top.arch.memory_cpu0.crossbar.ports.out_arbiter0: SEND @24 allow_slide=1 # 0x5565f39b6070,
21: top.arch.memory_cpu0.ports.in_crossbar_to_llc: RECEIVE SCHEDULED @24(1)  # 0x5565f39b6070

21: top.arch.memory_cpu0.crossbar.ports.out_arbiter0: SEND @25 allow_slide=1 # 0x5565f39b6070, prevDataSendCycle: 24, sendCycle: 25
21: top.arch.memory_cpu0.ports.in_crossbar_to_llc: RECEIVE SCHEDULED @25(2)  # 0x5565f39b6070
```

As you can see, the first sending at 21 is to be received at clock cycle 22. With `allow_slide` set, the second sending at 21 through the same port should be scheduled in 23. However, `computeNextAvailableCycleForSend` reports 24. Any succeeding sending through the same port is the previous one +1 which is correct again. The reason for the 24 is the call to `Clock::calculateClockCrossingDelay` which returns an incorrect delay for the send in the middle.

I have only a limited understanding, what happens in Sparta. Since I am leaving SyncPort now and trying to adapt `Clock`, I fear that I am going to break more than I fix. In addition, I am venturing into the area of clock domain crossing, which is again unknown terrain for me.

Since this PR disregards any clock domain crossing by just counting the packets in flight and adjusting the send delay accordingly, I would not recommend merging it. Consider, for instance, that the receiver could have a clock twice as fast as the sender and hence could accept 2 packets per clock cycle from the perspective of the sender. This PR does not take this into the account.

For me personally, counting the number of packets in flight works because I just have 1 clock domain. I would keep the PR to remember that the SyncPort has some issues with sliding packets. If allow_slide is removed, it renders this PR obsolete. Alternatively, someone who is comfortable with the inner workings of the SyncPort, should have a look into it.

alexanderFell: I have removed the access to the number of packets in flight and use `computeNextAvailableCycleForSend` with a few changes.
Again, Terms and Conditions apply: I am by no means an expert in the field. I see the advantage of using the method `computeNextAvailableCycleForSend` as it computes the next available cycle considering clock domain crossings. However, I do not have test cases to prove the correct behavior.

knute-sifive:
Finally have a chance to get back to this...
The clean and changes look good to me -- I'll approve this for now.
Although, I'm chin-scratching on this use-case:
```
if(outPorts->isReady()) {
    outPorts->send(message0, 1, false);
    outPorts->send(message1, 1, false);
    outPorts->send(message2, 1, false);
}
```
To me, you're requesting the SyncPort to send 3 different messages to the receiver _to be delivered at the same time_.  I know the SyncPort does not support this (`DataPort` however does), and I'm fine with that, but since SyncPort does **not** allow these 3 messages to be sent in same cycle, I would expect the modeler to be cognizant of this and change their code to this:

```
uint32_t delay = 1;
if(outPorts->isReady()) {
    outPorts->send(message0, delay++, false);
    outPorts->send(message1, delay++, false);
    outPorts->send(message2, delay, false);
}
```
The `allow_slide` tells the SyncPort "it's ok if you can't deliver the message on the next available cycle -- for whatever reason".

Need to ponder this a bit more...

knute-sifive (approved):

alexanderFell:
> To me, you're requesting the SyncPort to send 3 different messages to the receiver _to be delivered at the same time_. I know the SyncPort does not support this (`DataPort` however does), and I'm fine with that, but since SyncPort does **not** allow these 3 messages to be sent in same cycle, I would expect the modeler to be cognizant of this and change their code:

I totally agree with you here. The test bench is an extreme and dumb case and a quick way to construct and test a situation in which a collision occurs. In my project here, I never use `allow_slide`, since I never expect a collision to occur. I just came across `allow_slide` in the search for the other issues fixed in the recent PRs.

There are 2 ways to understand `allow_slide`:
1) Packets to be sent in the same clock cycle are allowing to slide (this PR). There is no limit to the number of packets to be scheduled for a specific cycle. However, packets are delivered sequentially and in-order.

2) There can be only 1 packet in flight. The purpose of `allow_slide` is that in case the port is not ready in the scheduled time slot (which is in the future), automatically the packet is rescheduled for delivery once the port is ready again. Once the packet has been delivered, the SyncOutPort can `send()` the next packet.

I think, depending on the point of view, `allow_slide` can be understood either way. This PR leans towards number 1 and reschedules packets in flight in case the port is not ready (2). The PR does not impose the constraint "one packet in flight" mentioned in 2).

At least, that is how I understood `allow_slide`. Please correct me, if I am wrong.

PR #368, opened 2022-09-30
Fix use of bitwise operator or with boolean operands

jackjchen:
Fixes a conda-forge clang14.04 warning that I found on osx-arm64:
`error: use of bitwise '|' with boolean operands [-Werror,-Wbitwise-instead-of-logical]`


klingaard (approved):
Man!  How long has THAT been there!

PR #369, opened 2022-10-03
Support for Arbitrary Report File Header Metadata

### LINKED ISSUE #279: Support for arbitrary header and fix warmup header value
This issue reports a bug and a new feature that provides an alternative, since the bug probably isn't fixable in sparta.

**The bug**
`my_sparta_sim --report reports/report_warmup.yaml --report-yaml-replacements  WARMUP_INSTS 29223  --report reports/time_series_crit_warmup.yaml --report-yaml-replacements WARMUP_INSTS 29223 TS_PERIOD 10 `
where
```
content:
  report:
    pattern: top
    def_file: reports/crit_stats.yaml
    dest_file: foo.csv
    format: csv
    trigger:
      start:  core_complex.core.retire.stats.insts_retired >= %WARMUP_INSTS%
      update-count: top.core_complex.core.retire.stats.insts_retired %TS_PERIOD%
```
results in a header of
```
# report="reports/crit_stats.yaml on top",start=39467,end=SIMULATION_END,report_format=csv
# enabled=none,period=10,after=29230,type=aligned_icount,counter=top.core_complex.core.retire.stats.insts_retired,terminate=none,warmup=18446744073709551615
```
Note the value of "warmup=" at the end of the 2nd line.  That's the problem.

I think that value comes from a bespoke usage of an earlier version of the reporting susbsystem that no longer makes sense.

**New Feature Request**
Add a generic way to provide report metadata for the header, as in:
```
content:
  report:
     header_metadata:
      stats_warmup:  %STATS_WARMUP%
      hostname: %HOSTNAME
```
where the `%%` tags would be replaced at simulation time via `--report-yaml-replacements`.
Ideally, the keys under "header_metadata" are not keywords, but arbitrary keys specified in the report description yaml.
These key value pairs shall be placed the report header, formatted appropriately for the json and csv report formats.

kathlenehurt-sifive:
Added mechanism to specify metadata to include in report files. Also removed all implicitly added report header metadata except for the counter location.

**Example**
Report definition file:
```
header_metadata:
  period: %TS_PERIOD%
  warmup: %WARMUP_INSTS%
content:
  report:
    pattern: top 
    def_file: reports/crit_stats.yaml
    dest_file: %OUT_BASE%_time_series_crit.csv
    format: csv 
    trigger:
      start:  'core_complex.core.retire.stats.insts_retired >= %WARMUP_INSTS% -> post.WarmupTrigger'
      update-count: top.core_complex.core.retire.stats.insts_retired %TS_PERIOD%
```
Report header:
```
# counter=top.core_complex.core.retire.stats.insts_retired,period=10,warmup=29223
```

Closes #279

### DIFF
--- a/sparta/src/ContextCounterTrigger.cpp
+++ b/sparta/src/ContextCounterTrigger.cpp
@@ -223,17 +223,8 @@ public:
 
     const CounterBase * getCounter() const
     {
-        if (internal_counters_.empty()) {
-            return nullptr;
-        }
-        if (internal_counters_.size() == 1) {
-            return internal_counters_[0];
-        }
-        throw SpartaException("You may not call the getCounter() method on this "
-                            "ContextCounterTrigger since it has ")
-            << internal_counters_.size()
-            << " internal counters, not just one.";
-        return nullptr;
+        return internal_counters_.empty() ? nullptr :
+            internal_counters_[0];
     }
 
     ~Impl() = default;
--- a/sparta/src/ContextCounterTrigger.cpp
+++ b/sparta/src/ContextCounterTrigger.cpp
@@ -223,8 +223,22 @@ public:
 
     const CounterBase * getCounter() const
     {
-        return internal_counters_.empty() ? nullptr :
-            internal_counters_[0];
+        // There is an assert in the constructor to ensure that internal_counters_ is never empty
+        if(internal_counters_.size() == 1) {
+            sparta_assert(false == internal_counters_[0]->isExpired(),
+                "Cannot getCounter on a ContextCounterTrigger because the referenced counter "
+                "has expired");
+            return internal_counters_[0];
+        }
+        else {
+            throw SpartaException("ContextCounterTrigger supports multiple counters. "
+                                  "Use \"getCounters()\" instead");
+        }
+    }
+
+    const Counters& getCounters() const
+    {
+        return internal_counters_;
     }
 
     ~Impl() = default;
@@ -321,9 +335,19 @@ bool ContextCounterTrigger::isTriggerReached_() const
     return impl_->isTriggerReached();
 }
 
+bool ContextCounterTrigger::supportsMultipleCounters() const
+{
+    return true;
+}
+
 const CounterBase * ContextCounterTrigger::getCounter() const
 {
     return impl_->getCounter();
+}
+
+const Counters& ContextCounterTrigger::getCounters() const
+{
+    return impl_->getCounters();
 }
 
 void ContextCounterTrigger::registerContextCounterCalcFunction(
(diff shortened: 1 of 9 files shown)


MichaelSchoenfelder-SiV:
Will this PR support this definition (header for WARMUP_INSTS, but no trigger)?
```
header_metadata:
  period: %TS_PERIOD%
  warmup: %WARMUP_INSTS%
content:
  report:
    pattern: top 
    def_file: reports/crit_stats.yaml
    dest_file: %OUT_BASE%_time_series_crit.csv
    format: csv 
    trigger:
      update-count: top.core_complex.core.retire.stats.insts_retired %TS_PERIOD%
```
It looks like it would, but that is my use case because then we could write a time series visualizer that showed the signal in both warmup and ROI, but also draw a vertical line at the beginning of the warmup period.

Another use for time series is for people to see where experiment A and B start to differ in time .  In that case, they want the time series to start at cycle zero, but would also like to know where the ROI starts.

kathlenehurt-sifive:
@MichaelSchoenfelder-SiV, your `update-count` example actually was not supported so I added metadata for the counters used for the start, stop and updated triggers. I'm not familiar enough with the Sparta expression triggers to know the full scope that needs to be supported so I only added metadata for counter-based triggers (only the first counter if the expression includes multiple counters).

sparta/src/ContextCounterTrigger.cpp
```diff
@@ -223,17 +223,8 @@ class ContextCounterTrigger::Impl
 
     const CounterBase * getCounter() const
     {
-        if (internal_counters_.empty()) {
-            return nullptr;
-        }
-        if (internal_counters_.size() == 1) {
-            return internal_counters_[0];
-        }
-        throw SpartaException("You may not call the getCounter() method on this "
```
kathlenehurt-sifive, line 232: @knute-sifive are you ok with me removing this exception? None of the other triggers have exceptions like this one so I didn't think it was functionally required.

knute-sifive: Did this assert fire on you?  Just curious.   Makes me scratch my head a bit -- if this assert ensures we only have 1 counter, then why is it a vector?

knute-sifive: Ok, I looked at this a bit in more detail -- memory's coming back.  The context counter can be a combination of multiple counters that attribute to an expression.  When "evaluated" it can be true (fire the trigger) or false.  `getCounter` really should return the array of counters involved; and each counter's value should be considered as part of the meta data.  I _think_ 😉

kathlenehurt-sifive: That's a good idea. I'll just change it to `getCounters()` that returns the whole vector. It was firing while I was trying to get the counter location to put in the header.

MichaelSchoenfelder-SiV:
I'd like to add 
```
header_metadata:
  period: %TS_PERIOD%
  warmup: %WARMUP_INSTS%
```
to all our existing time series reports, but we can do that in another PR.

kathlenehurt-sifive:
> WARMUP_INSTS

I added the header metadata section to the report for the "placeholders" test. I'll open a separate ticket to update the reports in our performance model.

PR #372, opened 2022-10-04
Add simple Scoreboard class to handle operand readiness

### LINKED ISSUE #263: A Simple Scoreboard to Handle Register Dependency
This issue describes a simple Scoreboard designed for in-order cores, which have the following characteristics:

- Instructions are in-order picked
- Instruction queue doesn't have **replay** capability
    - Instruction is directly deallocated from instruction queue after it is picked
    - When a speculative load-to-use fails to execute, the core flushes the consumer and re-fetches the instruction stream

An in-order scheduler has to know when the results of on-the-fly instructions are ready to use. So they can determine whether the issuing instructions can get data via bypassing network. 

This scoreboard can be just a list of counters, showing when the data of those registers become ready:
| Registers  | Cycles to Complete |
| - | - |
| x1  | 0 (Ready)  |
| x2  | 3 |
| ...  | ...  |
| x31  | 1 |

Major features
- void write(RegisterType reg, uint32_t cycle_to_write)
    - This method updates the scoreboard that the **reg** produces the result after **cycle_to_write**
- bool read(RegisterListType reg_list, uint32_t cycle_to_read)
    - This method checks whether those reading registers in **reg_list** are ready to use after **cycle_to_read**
- The Scoreboard is expected to automatically decrement non-zero counters

Unresolved
- Should it be a Sparta::Unit? Or just use others' event set?

aka2en:
Close #263


sparta/src/Scoreboard.cpp
```diff
@@ -0,0 +1,417 @@
...
+    }
+
+    void Scoreboard::set(const Scoreboard::RegisterBitMask & bits)
+    {
+        // Update registered scoreboards immediately
+        for(auto & sbvs : unit_id_to_scoreboard_views_)
+        {
+            for (auto * sbv : sbvs)
+            {
+                if(nullptr != sbv)
```
knute-sifive, line 156: I wonder (for performance) if the `sbvs` vector can just be a vector of valid scoreboards and we can remove this `if` statement.

sparta/src/Scoreboard.cpp
```diff
@@ -0,0 +1,417 @@
...
+            }
+        }
+    }
+
+    void Scoreboard::set(const Scoreboard::RegisterBitMask & bits, Scoreboard::UnitID producer)
+    {
+        sparta_assert(producer < forwarding_latencies_.size(),
+                      "could not find producer ID in forwarding_latencies table");
+
+        const auto & consumers = forwarding_latencies_[producer];
```
knute-sifive, line 169: It'd be cool if the `forwarding_latencies_` vector *only* contained valid scoreboards worth forwarding.  That way we can get rid of the `if` statment on line 173.  Same kinda logic for line 180.

knute-sifive (commented):
This is super awesome.  Wonder if it's better located directly in olympia than in sparta?  What are your thoughts?

aka2en:
> This is super awesome. Wonder if it's better located directly in olympia than in sparta? What are your thoughts?

I think sparta can include a really simple and easy-to-understand scoreboard and olympia can use it for now. Once the community has ideas to enhance the Scoreboard. I'm fine with either having new one in olympia or enhancing sparta's one.

sparta/sparta/resources/Scoreboard.hpp
```diff
@@ -0,0 +1,306 @@
...
+     * # The Scoreboard or "master" for each register file type (GPU,
+     *   FPR, Vector, etc).  Typically a Rename block is responsible
+     *   for setting/clearing the SB readiness.
+     *
+     * # The ScoreboardView is created by a Scheduling/Execution block
+     *   and is used to determine if an instruction is ready for
+     *   execution (all operands ready)
+     *
+     */
+    class Scoreboard : public sparta::Unit
```
knute-sifive, line 33: I'm thinking (for now) that this derive from `sparta::Resource`.  It's possible that a modeler might have a special logger/stat set for this and would prefer to pass them in.

kathlenehurt-sifive:
> > This is super awesome. Wonder if it's better located directly in olympia than in sparta? What are your thoughts?
> 
> I think sparta can include a really simple and easy-to-understand scoreboard and olympia can use it for now. Once the community has ideas to enhance the Scoreboard. I'm fine with either having new one in olympia or enhancing sparta's one.

I agree it should be in Sparta but we should make it more generic in the future since a scoreboard could be used for things other than physical registers.

sparta/src/Scoreboard.cpp
aka2en: Should work. Who wrote this stuff? :laughing:

sparta/src/Scoreboard.cpp
aka2en: Done. I needed an additional data structure to implement this.

sparta/sparta/resources/Scoreboard.hpp
aka2en: Question: how a modeler pass logger/stat in? Does modeler defined them in the derive class inheriting Scoreboard?

knute-sifive (approved):

sparta/sparta/resources/Scoreboard.hpp
knute-sifive: One way to do it is for the main scoreboard class to be dynamically constructed during the parent resource's constructor.  Since this is allowed in Sparta, the main constructor of the Scoreboard can either:
1. Look for it's parent's StatSet/Loggers 
2. If the parent _does not_ have a statset/logger, the SB creates its own and uses those

knute-sifive: The Scoreboard views should be instantiated in the constructor of the using class, which will have statset/logger already -- in theory.

sparta/src/Scoreboard.cpp
knute-sifive: I dunnnoooo, but I wouldn't trust him.

sparta/sparta/resources/Scoreboard.hpp
aka2en: @knute-sifive Gotcha. But I haven't figured out what's the benefit to use parent's StatSet/Loggers. Wouldn't it be clearer that having messages from Scoreboard instead of parent resource? l'll defer it until we run into that need.

knute-sifive: Couple of notes on the subject:
1. With the Scoreboard implementation as it is, the loggers are forced to `info`, `debug`, etc from `sparta::Unit`.  If I create a Scoreboard in my unit, the SB will be a sub-unit and logging enablement will _always include_ scoreboard output.  That would get verbose REALLY fast
2. If the SB has an API to support a separate logger, I can put the scoreboard on a "logger" scoreboard and separate out the messages.

PR #391, opened 2023-01-27
Replaced boost::filesystem::extension() calls with path::extension()

jerinjoy:
the macOS clang build was failing complaining about `boost::filesystem::extension()` being deprecated.

### DIFF
--- a/sparta/src/ReportDescriptor.cpp
+++ b/sparta/src/ReportDescriptor.cpp
@@ -15,6 +15,7 @@
 #include <boost/filesystem/convenience.hpp>
 #include <boost/filesystem/operations.hpp>
 #include <boost/iterator/iterator_facade.hpp>
+#include <filesystem>
 #include <fstream>
 #include <functional>
 #include <iostream>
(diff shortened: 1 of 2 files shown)


klingaard (approved):
LGTM, but might be missing 
`#include <filesystem>`

Thanks for doing this!

jerinjoy:
Fixed. clang didn't complain about this but gcc did.

klingaard:
Whelp, this is a can o' worms.  Looks like we need to move to 10.15 or later of MacOS in regression + a bizarre (and most likely unrelated) Scheduler timeout issue.  I can take a look at moving MacOS regression if you haven't already.

klingaard:
Another option is just address the original boost deprecation and continue using boost.

knute-sifive:
This problem has come up for us as well.  I've opened a new branch `knutel/jerinjoy-clang-filesystem-fix` with your changes merged.  Testing on a few machines I have access to before I start a PR with an update MacOS.

knute-sifive:
I'm going to close this PR in favor of PR #392

PR #393, opened 2023-01-31
Sparta perf improvements

knute-sifive:
These changes bumped performance of Sparta by roughly 10-20% on some simulators. 

- Added a Profile target
- Added a new class called `FastList` that is roughly 4x faster than `std::list`, but the user must provide an upper limit to the size
- Re-wrote the Scheduler's scheduling structures to not use std container types
- Added a new parameter to Scheduler::scheduleEvent that indicates if the event should be scheduled always or uniquely


sparta/sparta/events/PhasedPayloadEvent.hpp
```diff
@@ -10,13 +10,14 @@
 
 #pragma once
 
-#include <list>
+//#include <list>
```
kathlenehurt-sifive, line 13: Remove comment?

sparta/sparta/kernel/Scheduler.hpp
```diff
@@ -232,12 +263,26 @@ class Scheduler : public RootTreeNode
...
         }
 
-        Tick tick;     //!< The tick this quantum represents
-        Groups groups; //!< The list of firing groups. This is indexed by dag_group+1
+        void addEventIfNotScheduled(uint32_t firing_group, Scheduleable * scheduleable) {
+            sparta_assert(firing_group > 0);
+            sparta_assert(firing_group < groups.size());
+            auto & grp = groups[firing_group];
+            const auto grp_size = grp.size();
+            for(uint32_t idx = 0; idx < grp_size; ++idx) {
```
kathlenehurt-sifive, line 275: Could you use `std::find_if` instead?

sparta/sparta/events/PhasedPayloadEvent.hpp
knute-sifive: Done

sparta/sparta/kernel/Scheduler.hpp
knute-sifive: I could, but that'd add a second `if` statement.  Currently if there's a match, we bail out of this function right away.  If I change it to `find_if`, then I'd have to see if there were a match after the alg did all of the first `if` statements.  Good suggestion through, that'd clean it up.

kathlenehurt-sifive (approved):

Issue #395, opened 2023-02-03
Boost cmake: why use Boost_NO_BOOST_CMAKE ON on macOs?


peter-d:
(triggered by an Olympia build issue https://github.com/riscv-software-src/riscv-perf-model/issues/35)

I've noticed there's a bit of a non-standard way of finding boost for Sparta fro macOs: 
https://github.com/sparcians/map/blob/edc3b2a2285f6c36b2a5c8311b147649959b6926/sparta/cmake/sparta-config.cmake#L50

Which makes it need ```-DBoost_DIR=/opt/homebrew/Cellar/boost@1.76/1.76.0_3/lib/cmake/Boost-1.76.0``` on the command line to find the boost paths if they are not in a system default location.

But most other repos (like stf_lib, for example), to find boost with cmake, you define ```-DBOOST_ROOT=/opt/homebrew/Cellar/boost@1.76/1.76.0_3```

Removing that line,  on my system at least, boost is still found correctly with the standard ```-DBOOST_ROOT=<path>```, but then I am not using conda...

Is there a specific reason to keep this like that, or can the boost setup be made simpler (and more uniform)?

knute-sifive:
Getting boost working with MacOS/cmake a few years back was a nightmare.  Brew helps a bit, but there was something non-standard with how it was installed and how `cmake` interacted with it.  Perhaps this is now moot since we've moved _way_ past boost 1.65.  I think I took the attitude -- it ain't broke, so don't fix.

Now it's back to broke and not standard.  I'm ok with removing those constraints and trying again.

peter-d:
Thanks for the quick feedback. It's still not obvious - I am too worried that while this might now work for me it's not necessarily good for everyone, but I only have this system to test on...

peter-d:
We can close this now that #397 was merged a while ago.

PR #396, opened 2023-02-03
simpler boost detection on mac

peter-d:
Fix for #395, but not sure if there are no unwanted side effects, only tested on Ventura 13.1 on M1.

### DIFF
--- a/sparta/cmake/sparta-config.cmake
+++ b/sparta/cmake/sparta-config.cmake
@@ -45,12 +45,7 @@ else ()
     set (USING_CONDA OFF)
 endif ()
 
-
-if (APPLE AND NOT USING_CONDA)
-  set (CMAKE_CXX_COMPILER_VERSION 10.0)
-endif ()
-
-find_package (Boost 1.65.0 REQUIRED COMPONENTS ${_BOOST_COMPONENTS})
+find_package (Boost 1.76.0 REQUIRED COMPONENTS ${_BOOST_COMPONENTS})
 
 set (BUILD_SHARED_LIBS ${existing_build_shared})
 message (STATUS "Using BOOST ${Boost_VERSION_STRING}")
--- a/sparta/cmake/sparta-config.cmake
+++ b/sparta/cmake/sparta-config.cmake
@@ -27,11 +27,6 @@ endif()
 
 set (Boost_USE_STATIC_LIBS OFF)
 
-# BOOST_CMAKE logic for in versions before 1.72 to ask for the shared libraries is broken, you can only force it to use
-# them if you're building shared libs?  wtf?
-set (existing_build_shared ${BUILD_SHARED_LIBS})
-set (BUILD_SHARED_LIBS ON)
-
 execute_process (COMMAND ${CMAKE_CXX_COMPILER} --version OUTPUT_VARIABLE CXX_VERSION_STRING RESULT_VARIABLE rc)
 if (NOT rc EQUAL "0")
     message (FATAL_ERROR "could not run compiler command '${CMAKE_CXX_COMPILER} --version', rc=${rc}")
@@ -47,7 +42,6 @@ endif ()
 
 find_package (Boost 1.76.0 REQUIRED COMPONENTS ${_BOOST_COMPONENTS})
 
-set (BUILD_SHARED_LIBS ${existing_build_shared})
 message (STATUS "Using BOOST ${Boost_VERSION_STRING}")
 
 # Find YAML CPP
--- a/sparta/cmake/sparta-config.cmake
+++ b/sparta/cmake/sparta-config.cmake
@@ -47,12 +47,10 @@ endif ()
 
 
 if (APPLE AND NOT USING_CONDA)
-  set (Boost_NO_BOOST_CMAKE ON)
   set (CMAKE_CXX_COMPILER_VERSION 10.0)
-  find_package (Boost 1.65.0 REQUIRED HINTS /usr/local/Cellar/boost/* COMPONENTS ${_BOOST_COMPONENTS})
-else ()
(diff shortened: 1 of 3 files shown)


sparta/cmake/sparta-config.cmake
```diff
@@ -47,13 +47,11 @@ endif ()
...
 
 if (APPLE AND NOT USING_CONDA)
-  set (Boost_NO_BOOST_CMAKE ON)
   set (CMAKE_CXX_COMPILER_VERSION 10.0)
-  find_package (Boost 1.65.0 REQUIRED HINTS /usr/local/Cellar/boost/* COMPONENTS ${_BOOST_COMPONENTS})
-else ()
-  find_package (Boost 1.65.0 REQUIRED COMPONENTS ${_BOOST_COMPONENTS})
 endif ()
 
+find_package (Boost 1.65.0 REQUIRED COMPONENTS ${_BOOST_COMPONENTS})
```
knute-sifive, line 53: If you don't mind, let's move this to 1.76

peter-d:
I don’t mind! Do you know why the CXX compiler version is set to 10 here? If I can take that out, it becomes even simpler: no more delta vs non-Mac systems. Op 3 feb. 2023 om 17:38 heeft Knute Lingaard ***@***.***> het volgende geschreven:﻿
@knute-sifive commented on this pull request.



In sparta/cmake/sparta-config.cmake:
>  endif ()
 
+find_package (Boost 1.65.0 REQUIRED COMPONENTS ${_BOOST_COMPONENTS})

If you don't mind, let's move this to 1.76

knute-sifive:
I tested on a few machines with ALL of that code removed -- and it worked, so yes!  Please remove it all!  Thanks!

peter-d:
Hmm - the cleaner include path set up seems to break in CI, though it works in the systems I have access too (MacOS and WSL2 Ubuntu). Not sure how to debug that without access to a system where it fails.

@knute-sifive  I also noticed my WSL Ubuntu does not yet have Boost 1.76, only 1.74. Not sure which version you like to pin to? I now updated the cmake to require 1.76, but that might be overly aggressive for some distros... Not sure how this lines up with the systems you typically run on.

knute-sifive:
I'm ok with an (slightly) older of boost if that makes standard environments work.  I'm looking at the CI failures and it appears that an include path to `simdb` was dropped.  Odd.

knute-sifive:
I'm building now...

I like the direction this PR is going, but one concern I have is with regards to the users of sparta.  The purpose of the `sparta/cmake/sparta-config.cmake` file is so that modelers that use Sparta don't have to explicitly set up `SYSTEM` includes or library paths/targets based on what Sparta requires.  Instead, a modeler just includes this `cmake` file and dependencies are automatically set up.

With that hope said, I think the real issue is that the `sparta/cmake/sparta-config.cmake` is incomplete.  In other words, it finds packages for you, sets up library paths/linking, but it did _not_ set up include paths.  This is missing.

I have a branch where I address that.  Can you take a look at it and merge it with yours?  I'm not seeing the `simdb` issue...

https://github.com/sparcians/map/tree/knutel/peter-d_cmake_include_fixes

knute-sifive:
Ok, I can reproduce the failure locally.  Apparently I had `simdb` installed locally and the paths were automatically found!  I found `simdb` in `/usr/local/include`.  Guess I was playing around with installation packages at one time.

knute-sifive:
Fixed the `simdb` issue on my other branch.

peter-d:
> I'm building now...
> 
> I like the direction this PR is going, but one concern I have is with regards to the users of sparta. The purpose of the `sparta/cmake/sparta-config.cmake` file is so that modelers that use Sparta don't have to explicitly set up `SYSTEM` includes or library paths/targets based on what Sparta requires. Instead, a modeler just includes this `cmake` file and dependencies are automatically set up.

I fully agree - it should be as simple as an include and (maybe) a find_package (which will allow you down the line to select versions If needed)

We actually made an internal ```FindSparta.cmake``` that does pretty much this, and will set up Sparta and all dependencies, nicely in ```Sparta::``` cmake namespace, so you can also easily just add Sparta to ```target_link_libraries``` statements and the like. I'll need to check if I am allowed to share that publicly - should not be an issue, normally, but just need to be sure. If you like, I can contribute that, either directly or as inspiration. 

> 
> With that hope said, I think the real issue is that the `sparta/cmake/sparta-config.cmake` is incomplete. In other words, it finds packages for you, sets up library paths/linking, but it did _not_ set up include paths. This is missing.

Correct, this is now indeed spread over 2 files, part in the config, part in the ```CMakeLists.txt`

> 
> I have a branch where I address that. Can you take a look at it and merge it with yours? I'm not seeing the `simdb` issue...
> 
> https://github.com/sparcians/map/tree/knutel/peter-d_cmake_include_fixes

I'll have a look right away!

knute-sifive:
> We actually made an internal FindSparta.cmake that does pretty much this, and will set up Sparta and all dependencies, nicely in Sparta:: cmake namespace, so you can also easily just add Sparta to target_link_libraries statements and the like. I'll need to check if I am allowed to share that publicly - should not be an issue, normally, but just need to be sure. If you like, I can contribute that, either directly or as inspiration.

🤤 

That would be awesome!!

knute-sifive:
I'm slowly picking away the regression failures in the other PR and I think (🤞🏻 ) I got it working.

But a fundamental question on the `cmake` philosophy: Sparta is a package that _depends_ on many other packages.  How do you ensure that an end user (the modeler) is using the same packages that Sparta was built with?  

As an example, Sparta was built with Boost 1.71.  But the modeler is using Boost 1.80.  What magic keeps everything in sync?  🤔

peter-d:
Superseded by #397

PR #397, opened 2023-02-06
Knutel/peter d cmake include fixes

knute-sifive:


peter-d (commented):
Much cleaner already, nice - but a few remarks in the comments.

Also I was thinking: it's not very easy to see based on the config file what I need to do if I build a project to also link against libsparta or libsimdb...

It would be the most user friendly if you can just add Sparta to the ```target_link_libraries``` statement.

sparta/cmake/sparta-config.cmake
```diff
@@ -45,39 +38,37 @@ else ()
...
-else ()
-  find_package (Boost 1.65.0 REQUIRED COMPONENTS ${_BOOST_COMPONENTS})
-endif ()
-
-set (BUILD_SHARED_LIBS ${existing_build_shared})
+# Find Boost
+set (Boost_USE_STATIC_LIBS OFF)
+find_package (Boost 1.76.0 REQUIRED COMPONENTS ${_BOOST_COMPONENTS})
 message (STATUS "Using BOOST ${Boost_VERSION_STRING}")
+include_directories (SYSTEM ${Boost_INCLUDE_DIRS})
```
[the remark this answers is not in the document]
knute-sifive: Ah, I see why you removed them.  Ok.  Let me look at your PR from this perspective then.

sparta/CMakeLists.txt
```diff
@@ -105,6 +88,12 @@ if (COMPILE_WITH_PYTHON)
     python/sparta_support/Completer.cpp)
 endif ()
 
+add_library (sparta ${SourceCppFiles})
+
+# Add local includes
+target_include_directories (sparta PRIVATE "./")
```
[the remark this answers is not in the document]
knute-sifive: Good point.

peter-d:
And we did still break the examples in CI, though they build in this branch for me locally...

knute-sifive:
> It would be the most user friendly if you can just add Sparta to the target_link_libraries statement.

Agreed!

sparta/CMakeLists.txt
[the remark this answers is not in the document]
knute-sifive: But then again, this is in the private CMakeLists.txt of Sparta...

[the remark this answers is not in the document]
knute-sifive: Ah, but it does not transfer to the regression if PRIVATE.  Good catch.

peter-d:
Hi Knute, 

In trying to get my cmake file to find sparta up and running and testing it on building Olympia, I've noticed that Olympia depends on a few parts of the sparta tests folder:
- TestingMacros.cmake which has convenience macros
- but also a bunch of (test?) C++ code from sparta/tests

I was currently moving to a setup where make install installs:
- include files
- libsparta.a and libsimdb.a
- and also the cmake files that other projects need to link against sparta

Should the test code be installed as well? Right now ```make install``` does not include these. Or is Olympia an exceptional case?

That said: I've come to a point where it can Olympia can find sparta  after a make install, generate all makefiles but then fails as it cannot include files from map/sparta/test (in my case I need to point it to the right folder with a command line arg, as it is not installed in a system wide location like /usr/local or such)

knute-sifive:
> * TestingMacros.cmake which has convenience macros
> * but also a bunch of (test?) C++ code from sparta/tests

Olympia depends on the tests?  Odd (and sounds broken).

> I was currently moving to a setup where make install installs:
> 
> * include files
> * libsparta.a and libsimdb.a
> * and also the cmake files that other projects need to link against sparta
> 
> Should the test code be installed as well? Right now `make install` does not include these. Or is Olympia an exceptional case?

My original intent was just a dependency on the TestingMacros.cmake.  I'm wondering if the install target could copy that file over...  But your previous statement has be concerned.

> That said: I've come to a point where it can Olympia can find sparta after a make install, generate all makefiles but then fails as it cannot include files from map/sparta/test (in my case I need to point it to the right folder with a command line arg, as it is not installed in a system wide location like /usr/local or such)

That's AWESOME!  And a bummer on the testing.  None of this is set in stone.  If this becomes too complex, other routes can be explored.   

Thanks for helping with this.

peter-d:
> My original intent was just a dependency on the TestingMacros.cmake. I'm wondering if the install target could copy that file over... But your previous statement has be concerned.

That’s what I did now. It sits next to FindSparta.cmake and gets included when you find_package(Sparta). But turns out that’s not enough to get Olympia to build. 

I’m not at my computer tonight but will do some more digging on the test dependencies once I can.

peter-d:
The culprit is Dispatch, which instantiates a ```WeightedContextCounter``` which is included from sparta source tree: ```test/ContextCounter/WeightedContextCounter.hpp```: https://github.com/riscv-software-src/riscv-perf-model/blob/697fdc2a950be59e67a78a53220f3eb2a0c03e39/core/Dispatch.hpp#L19

A quick grep shows that this is the only place where this happens.

Which then of course gives problems:
```
[ 41%] Building CXX object core/CMakeFiles/core.dir/Dispatch.cpp.o
cd /Users/pdback/src/peter-d/riscv-perf-model/release/core && /Library/Developer/CommandLineTools/usr/bin/c++ -DBOOST_ALL_NO_LIB -DBOOST_CHRONO_DYN_LINK -DBOOST_FILESYSTEM_DYN_LINK -DBOOST_PROGRAM_OPTIONS_DYN_LINK -DBOOST_SERIALIZATION_DYN_LINK -DBOOST_TIMER_DYN_LINK -DOLYMPIA_VERSION=\"v0.1.0\" -I/Users/pdback/src/peter-d/riscv-perf-model/core -I/Users/pdback/src/peter-d/riscv-perf-model/mss -isystem /opt/homebrew/include -isystem /Users/pdback/src/peter-d/riscv-perf-model/mavis -isystem /Users/pdback/src/peter-d/riscv-perf-model/stf_lib -isystem /simdb/include -isystem /opt/homebrew/Cellar/boost@1.76/1.76.0_3/include -isystem /Users/pdback/opt/sparta/include -isystem /opt/homebrew/Cellar/hdf5/1.12.2_2/include -isystem /opt/homebrew/opt/libaec/include -O3 -DNDEBUG -arch arm64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX13.1.sdk -Wall -Wextra -Winline -Winit-self -Wno-unused-function -Wuninitialized -Wno-sequence-point -Wno-inline -Wno-unknown-pragmas -Woverloaded-virtual -Wno-unused-parameter -Wno-missing-field-initializers -pipe -O3 -mllvm -inline-threshold=1024 -fomit-frame-pointer -flto=thin -std=c++17 -fPIC -Wconversion -std=gnu++17 -MD -MT core/CMakeFiles/core.dir/Dispatch.cpp.o -MF CMakeFiles/core.dir/Dispatch.cpp.o.d -o CMakeFiles/core.dir/Dispatch.cpp.o -c /Users/pdback/src/peter-d/riscv-perf-model/core/Dispatch.cpp
In file included from /Users/pdback/src/peter-d/riscv-perf-model/core/Dispatch.cpp:4:
In file included from /Users/pdback/src/peter-d/riscv-perf-model/core/CoreUtils.hpp:10:
In file included from /Users/pdback/src/peter-d/riscv-perf-model/core/CPUTopology.hpp:9:
In file included from /Users/pdback/src/peter-d/riscv-perf-model/core/CPUFactories.hpp:11:
/Users/pdback/src/peter-d/riscv-perf-model/core/Dispatch.hpp:19:10: fatal error: 'test/ContextCounter/WeightedContextCounter.hpp' file not found
#include "test/ContextCounter/WeightedContextCounter.hpp"
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
make[2]: *** [core/CMakeFiles/core.dir/Dispatch.cpp.o] Error 1
make[1]: *** [core/CMakeFiles/core.dir/all] Error 2
make: *** [all] Error 2
```

Looks to me like either that example weighted counter is useful enough to be part of sparta proper (there might be others like it?), or we should add the file to Olympia as a utility defined and used locally there.

peter-d:
I've pushed what I have for now - FindSparta.cmake courtesy of my colleague @tomhaber, who put this together.

A few things we should fix here before this is ready to merge:
- I'd like to use the find_package construct for the tests or at least a test, so it's covered in regression and we can see it break there rather than in projects depending on sparta.
- This is now a manually generated cmake file, but the more modern, proper way is to use [```install(EXPORT ...)```](https://cmake.org/cmake/help/latest/command/install.html#installing-exports) which will likely be more robust

peter-d:
So, ignoring the dependency on the counter from test, I also see Olympia needs ```map/sparta/cache``` code, which is also not installed in the default ```make install```, which only copies over ```map/sparta/sparta``` to the installation destination.

What's the intent there? is cache supposed to be a fixed part of sparta? Or is it more like stf_lib and mavis, a set of utilities you can optionally use on top of/with sparta?

Depending on that answer, it's relatively easy to install it, but question is if we want to/should.

knute-sifive:
> Looks to me like either that example weighted counter is useful enough to be part of sparta proper

Ugh, I have NO idea how that happened.  Yes, it can be made proper esp since it has it own test.  I'll work on that.

knute-sifive:
> I've pushed what I have for now - FindSparta.cmake courtesy of my colleague @tomhaber, who put this together.
> 
> A few things we should fix here before this is ready to merge:
> 
>     * I'd like to use the find_package construct for the tests or at least a test, so it's covered in regression and we can see it break there rather than in projects depending on sparta.
> 
>     * This is now a manually generated cmake file, but the more modern, proper way is to use [`install(EXPORT ...)`](https://cmake.org/cmake/help/latest/command/install.html#installing-exports) which will likely be more robust
> 
>     * We should add/update something in the README or docs on how to use sparta in other projects.

All great ideas and I'm in complete agreement.  Let me know how I can help.

knute-sifive:
> So, ignoring the dependency on the counter from test, I also see Olympia needs `map/sparta/cache` code, which is also not installed in the default `make install`, which only copies over `map/sparta/sparta` to the installation destination.
> 
> What's the intent there? is cache supposed to be a fixed part of sparta? Or is it more like stf_lib and mavis, a set of utilities you can optionally use on top of/with sparta?
> 
> Depending on that answer, it's relatively easy to install it, but question is if we want to/should.

For now, yes, it's part of Sparta.  I'm working with a colleague here at SiFive to make a more proper cache repository in `map` (or higher) and enable the API to work with other cache simulators out there in the wild (Gem5 Ruby stuff for example).  Will be some time before that happens, though.

knute-sifive:
> ps this is quickly becoming a very deep rabbit hole ;-)

Yep.  🥲   But will be beautiful when done.  I seriously do appreciate the effort you and team are putting into this.

knute-sifive:
Curious to see the changes you made to Olympia to get this working.  I tried on my own and got pretty far, but I'm hitting some snags (lack of understanding on my end).

peter-d:
> Curious to see the changes you made to Olympia to get this working. I tried on my own and got pretty far, but I'm hitting some snags (lack of understanding on my end).

I've pushed my changes to the Olympia pr. What's there won't build, I have on my local machine commented out the missing counter code in Dispatch, but did not push that to Olympia - better just wait until it gets included in sparta proper first, so we just fix it properly there right away.

peter-d:
> For now, yes, it's part of Sparta. I'm working with a colleague here at SiFive to make a more proper cache repository in `map` (or higher) and enable the API to work with other cache simulators out there in the wild (Gem5 Ruby stuff for example). Will be some time before that happens, though.

I'll just install it next to sparta for now (just like it is next to the sparta folder in the source tree) - you can change that when it's ready to grow and move elsewhere.

peter-d:
> Argos uses this files from Sparta to build:

Managed to work a bit further on this in between everything else:
- I've used the FindSparta.cmake to now let everything in the helios folder find the sparta dependencies it needs. 
- I've also centralised all the testing macros cmake (both for sparta and simdb - a lot of duplication there btw!) (that's a bit of a touch-all on all CMakeLists.txt in the test folders, but a good cleanup)

Todo still: pipeviewer install: I'm struggling a bit with the python + .so files install, in order to put everything there in such a way that python will always find it. The cleaner way likely is to wrap everything in the ```python setup.py``` code and use that to generate a wheel that includes the binaries, which you can then install with ```pip install``` But I don't know enough on python setuptools to know how to get that done...

See my pushes, let's see if CI likes it...

peter-d:
Hopefully fixed CI, that's going to be it for me this week on this PR (but to be continued ;-) )

peter-d:
I have almost fixed the python pipe_view install as well, but there's a hard-coder path for transactiondb: https://github.com/sparcians/map/blob/ef6580d4b90826d7b3eb2a9dae972da74397c2f9/helios/pipeViewer/pipe_view/model/database.py#L15

Are you ok to restructure this to only use either
-  a relative path if you run in-tree? (might be tricky and will be depending on the name you pick for the build folder in cmake)
- somewhere in LD_LIBRARY_PATH or any other way to find it?

knute-sifive:
Hey @peter-d, a quick heads up regarding PR #406: not sure this will impact this work (don't believe it will), but we needed to move basic compiler requirements forward specifically to support `filesystem`.

peter-d:
> Hey @peter-d, a quick heads up regarding PR #406: not sure this will impact this work (don't believe it will), but we needed to move basic compiler requirements forward specifically to support `filesystem`.

Thank you for the head's up! It does not look like it interferes, but looks can be deceiving of course. I am not running a conda based set up here but homebrew python. So what works for me might break for conda based installs. (and hence also CI. At least we'll see it then)

I am currently revamping the whole setup.py / CMakeLists.txt situation for pipe view / Argos. The approach I am following is this one: https://github.com/kmhsonnenkind/cmake-cython-example - less fancy manipulation and dependency detections in setup.py and maximally pass all required build params/ibs/includes etc from CMakeLists (which already knows all of them) wxWidgets is not playing very nice so it won't look as clean as I want, but hopefully much cleaner than it currently is. 

It's a bit of a maze to navigate though, but looks like I am getting quite close to get it to work at least locally here...

peter-d:
Hi @knute-sifive I've pushed a major touch-all commit https://github.com/sparcians/map/pull/397/commits/3cc1cf088e3135229d536f3df23e185caac281f4 to get the python code cleaned up. I had to do a massive cleanup/reorg of where the files are and how we build it all.

I have no guarantee that this will not break in conda env, I've tested this locally with homebrew python, but CI will tell, I guess.

To summarise how it works now:
1. you configure with like ```make -S . -B release -DCMAKE_BUILD_TYPE=Release -DPython3_ROOT_DIR=$(brew --prefix)/Frameworks/Python.framework/Versions/3.10/ -DBOOST_ROOT=$(brew --prefix)/opt/boost@1.76  -DCMAKE_INSTALL_PREFIX=$HOME/opt/sparta```
   - Python3 and Boost are optional if you don't provide them, it will take whatever find_package finds first.
   - CMAKE_INSTALL_PREFIX is where sparta will be installed (if not provided system default will be used, I guess /usr/local or something like that
3. Then build: ```make -C release -j```
4.  Then install: ```make -C release install``` (if 2. not done then it will build anyways)
    -   This will install the libs somewhere in CMAKE_INSTALL_PREFIX/{lib, include} etc. 
    -   There will also be cmake scripts such that find_package(Sparta) works
    -   This will also install Argos (and all the C++ modules built) into the site-packages of whatever python was found by cmake (I've set it up such that virtual envs and conda take preference over what is in the path, but the -DPython3 option overrides all of that)

After this, I can build Olympia with a command like this:
```
cmake -S . -B release -DCMAKE_BUILD_TYPE=Release -DSPARTA_SEARCH_DIR=$HOME/opt/sparta  -DBOOST_ROOT=$(brew --prefix)/opt/boost@1.76
```
 (You'll need to point it to where sparta is if it cannot be found automatically, if installed in a system default location, it should not be needed)

And then, you don't need to do ```python <wherever Argos source .py is>``` to run pipe viewer. Instead, since it gets installed in a bin folder in your path, you can just do ```argos -d pipeout_1K -l ../layouts/small_core.alf```

**TODO**
(to be discussed if we want some/all of this before we merge or if we can tackle some separately)
- review and clean up
- python: how much of the path and import magic can we remove? There was a lot of python code to magically change python paths to find everything. I removed most of it, but kept ENV variable based overrides here and there, but _in theory_, none of that is needed anymore, setuptools will have taken care of that for us.
- Right now cmake installs a hand baked FindSparta.cmake file, but it can be upgraded to a more modern install(EXPORT ...) instead

peter-d:
So I guess this means this is now functionally complete. (and yay even CI has passed! 😄 )

But as you can see: many many files touched and things reorged in the repo. I am worried this might break things in unexpected ways for users, that I cannot see myself...

helios/pipeViewer/CMakeLists.txt
```diff
@@ -1,31 +1,114 @@
...
+#                   COMMAND ${CMAKE_COMMAND} -E touch ${PY_OUTPUT}
+#                   DEPENDS ${SETUP_PY} "${CMAKE_CURRENT_SOURCE_DIR}/pipe_view/core/src/core.pyx"
+#                   )
+#add_custom_target(pipe_view_py ALL DEPENDS "${PY_OUTPUT}")
+##
+#add_subdirectory(transactiondb)
+#add_subdirectory(transactionsearch)
+#add_subdirectory(argos_dumper)
+#
+#install(DIRECTORY pipe_view/ DESTINATION pipe_view PATTERN "__pycache__" EXCLUDE)
```
knute-sifive, line 114: I'm ok with removing older code (even if it were there for reference).

knute-sifive:
Wow!  This is great!  I'm loving this change:
>  since it gets installed in a bin folder in your path, you can just do argos -d pipeout_1K -l ../layouts/small_core.alf

Nice! It appears that `conda` is not unhappy with these changes since (as you put it), CI passed (which uses Conda).  

Curious, what do you mean by this question?
> What do we do with the cache files? Right now I install it next to the other sparta includes and libs, but this might not be what you want
Which cache files are you referring to?

I'm merging this branch with master and testing with my Conda environment on my Mac and I appear to have a clean build.  Regression is still running, but I'm confident.   I'll push the merge when complete.

helios/pipeViewer/pipe_view/model/database.py
```diff
@@ -12,36 +12,11 @@
...
-if env_var is None:
-    # Try to find the transaction module in the Helios release dir
-    added_path = os.path.dirname(__file__) + \
-        "/../../../../release/helios/pipeViewer/transactiondb/lib"
-    added_path = os.path.abspath(added_path)
-    if not os.path.isdir(added_path):
-        error('Argos cannot find the transactiondb directory: %s', added_path)
-        sys.exit(1)
-else:
-    added_path = os.environ.get(__MODULE_ENV_VAR_NAME, os.getcwd())
```
knute-sifive, line 27: @bdutro , with Peter's changes (which includes a proper install of Argos/Sparta), this goes away.  @MichaelSchoenfelder-SiV, we should chat about how to take advantage of this.

bdutro: This should be fine, worst case we can always tweak `PYTHONPATH` with a wrapper script.

bdutro: Nevermind, just saw that we're doing a relative import now. That should make it even easier.

knute-sifive:
So I did a `make install` on my Mac and it installed beautifully (using `conda`) with one exception:  I had to make 2 changes to `shortcut_help.py` due to these errors:
```
    __CTRL_KEY = 'Command' if is_mac_os() else 'CTRL'
TypeError: 'staticmethod' object is not callable

    __MESSAGE = __gen_message(__SHORTCUT_ITEMS)
TypeError: 'staticmethod' object is not callable
```
I just pushed these changes.  Do you mind testing them?  Basically, `make install` followed by `argos` should do the trick.

knute-sifive:
Also, on the Mac, with `conda`, `pythonw` is required:

```
cd release/sparta/sparta/example/CoreModel
pythonw `which argos` -d pipeout -l cpu_layout.alf 
```

bdutro:
The changes to the Python imports break `mypy` and `flake8`:

```
$ mypy argos.py       
model/element.py: error: Source file found twice under different module names: "pipe_view.model.element" and "model.element"
model/element.py: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules for more info
model/element.py: note: Common resolutions include: a) adding `__init__.py` somewhere, b) using `--explicit-package-bases` or adjusting MYPYPATH
Found 1 error in 1 file (errors prevented further checking)
$ flake8 $(git ls-files '*.py')       
argos.py:41:80: E501 line too long (81 > 79 characters)
argos.py:49:1: E302 expected 2 blank lines, found 1
argos.py:465:1: E305 expected 2 blank lines after class or function definition, found 1
gui/dialogs/watchlist_dialog.py:4:43: E127 continuation line over-indented for visual indent
model/database.py:5:1: F401 'os' imported but unused
model/database.py:6:1: F401 'sys' imported but unused
model/element.py:28:1: F401 '..gui.autocoloring' imported but unused
model/element.py:1025:70: F821 undefined name 'gui'
model/element.py:1028:78: F821 undefined name 'gui'
model/element.py:1030:75: F821 undefined name 'gui'
model/element.py:1032:49: F821 undefined name 'gui'
```

bdutro:
@knute-sifive I tweaked your fix to just move those functions outside the class.

bdutro:
I fixed the flake8 warnings, but mypy is still an issue.

bdutro:
Getting this error trying to build on my Linux box:

```
g++ -shared -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now -flto=auto -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now -flto=auto -pthread -lwx_gtk3u_core-3.2 -lwx_baseu-3.2 -DWXUSINGDLL -D__WXGTK3__ -D__WXGTK__ build/temp.linux-x86_64-cpython-310/pipe_view/transactiondb/src/transactiondb.o -L/usr/lib -L/home/brett/sifive/map_native/release/sparta -L/home/brett/sifive/map_native/release/sparta/simdb -L/usr/lib -lsparta -lsimdb -lhdf5 -lsqlite3 -o build/lib.linux-x86_64-cpython-310/pipe_view/transactiondb.cpython-310-x86_64-linux-gnu.so
/usr/bin/ld: /home/brett/sifive/map_native/release/sparta/libsparta.a(sparta.cpp.o): warning: relocation against `_ZN6sparta13SleeperThread15sleeper_thread_E' in read-only section `.text.startup'
/usr/bin/ld: /home/brett/sifive/map_native/release/sparta/libsparta.a(SpartaException.cpp.o): relocation R_X86_64_PC32 against symbol `_ZTVN6sparta15SpartaExceptionE' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: bad value
collect2: error: ld returned 1 exit status
error: command '/usr/bin/g++' failed with exit code 1
make[2]: *** [helios/pipeViewer/CMakeFiles/pipeView.dir/build.make:75: helios/pipeViewer/cython.stamp] Error 1
make[1]: *** [CMakeFiles/Makefile2:5899: helios/pipeViewer/CMakeFiles/pipeView.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
```

I specified `CC=clang CXX=clang++` when I ran `cmake`.

bdutro:
The initial compile for transactiondb.cpp also uses `gcc`:

```
gcc -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -march=x86-64 -mtune=generic -O3 -pipe -fno-plt -fexceptions -Wp,-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security -fstack-clash-protection -fcf-protection -g -ffile-prefix-map=/build/python/src=/usr/src/debug/python -flto=auto -ffat-lto-objects -march=x86-64 -mtune=generic -O3 -pipe -fno-plt -fexceptions -Wp,-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security -fstack-clash-protection -fcf-protection -g -ffile-prefix-map=/build/python/src=/usr/src/debug/python -flto=auto -march=x86-64 -mtune=generic -O3 -pipe -fno-plt -fexceptions -Wp,-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security -fstack-clash-protection -fcf-protection -g -ffile-prefix-map=/build/python/src=/usr/src/debug/python -flto=auto -DWXUSINGDLL -D__WXGTK3__ -D__WXGTK__ -fPIC -Ipipe_view/transactiondb/src -Ipipe_view/core/src -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/home/brett/sifive/map_native/sparta/./ -I/home/brett/sifive/map_native/sparta/simdb/include -I/usr/lib/python3.10/site-packages/wx/include -I/usr/lib/wx/include/gtk3-unicode-3.2 -I/usr/include/wx-3.2 -I/usr/include/python3.10 -c pipe_view/transactiondb/src/transactiondb.cpp -o build/temp.linux-x86_64-cpython-310/pipe_view/transactiondb/src/transactiondb.o --std=c++17 -Wno-cast-qual -Wno-deprecated-declarations -Wno-strict-aliasing -Wall -Wpedantic
In file included from pipe_view/transactiondb/src/Reader.hpp:26,
                 from pipe_view/transactiondb/src/TransactionDatabaseInterface.hpp:24,
                 from pipe_view/transactiondb/src/transactiondb.cpp:818:
/home/brett/sifive/map_native/sparta/./sparta/utils/LexicalCast.hpp: In function ‘T sparta::lexicalCast(const std::string&, uint32_t)’:
/home/brett/sifive/map_native/sparta/./sparta/utils/LexicalCast.hpp:85:25: warning: the address of ‘ss’ will never be NULL [-Waddress]
   85 |         if(((void*)&ss) != nullptr){
      |            ~~~~~~~~~~~~~^~~~~~~~~~
/home/brett/sifive/map_native/sparta/./sparta/utils/LexicalCast.hpp:82:27: note: ‘ss’ declared here
   82 |         std::stringstream ss;
      |                           ^~
```

(Side note, is that check at `LexicalCast.hpp:85` doing what we think it should be doing?)

peter-d:
> Curious, what do you mean by this question?

It's related to this comment: https://github.com/sparcians/map/pull/397#issuecomment-1424930971 
Right now the cache files are installed as well, but they sit next to sparta includes. I guess that's what it should be for now?

peter-d:
> So I did a `make install` on my Mac and it installed beautifully (using `conda`) with one exception: I had to make 2 changes to `shortcut_help.py` due to these errors:
> 
> ```
>     __CTRL_KEY = 'Command' if is_mac_os() else 'CTRL'
> TypeError: 'staticmethod' object is not callable
> 
>     __MESSAGE = __gen_message(__SHORTCUT_ITEMS)
> TypeError: 'staticmethod' object is not callable
> ```
> 
> I just pushed these changes. Do you mind testing them? Basically, `make install` followed by `argos` should do the trick.

That fix broke for me, the ```__``` makes it outside of scope where you want to use it in the class. Changed this to a single ```_``` and that works.

peter-d:
> Getting this error trying to build on my Linux box:
> 
> ```
> g++ -shared -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now -flto=auto -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now -flto=auto -pthread -lwx_gtk3u_core-3.2 -lwx_baseu-3.2 -DWXUSINGDLL -D__WXGTK3__ -D__WXGTK__ build/temp.linux-x86_64-cpython-310/pipe_view/transactiondb/src/transactiondb.o -L/usr/lib -L/home/brett/sifive/map_native/release/sparta -L/home/brett/sifive/map_native/release/sparta/simdb -L/usr/lib -lsparta -lsimdb -lhdf5 -lsqlite3 -o build/lib.linux-x86_64-cpython-310/pipe_view/transactiondb.cpython-310-x86_64-linux-gnu.so
> /usr/bin/ld: /home/brett/sifive/map_native/release/sparta/libsparta.a(sparta.cpp.o): warning: relocation against `_ZN6sparta13SleeperThread15sleeper_thread_E' in read-only section `.text.startup'
> /usr/bin/ld: /home/brett/sifive/map_native/release/sparta/libsparta.a(SpartaException.cpp.o): relocation R_X86_64_PC32 against symbol `_ZTVN6sparta15SpartaExceptionE' can not be used when making a shared object; recompile with -fPIC
> /usr/bin/ld: final link failed: bad value
> collect2: error: ld returned 1 exit status
> error: command '/usr/bin/g++' failed with exit code 1
> make[2]: *** [helios/pipeViewer/CMakeFiles/pipeView.dir/build.make:75: helios/pipeViewer/cython.stamp] Error 1
> make[1]: *** [CMakeFiles/Makefile2:5899: helios/pipeViewer/CMakeFiles/pipeView.dir/all] Error 2
> make: *** [Makefile:136: all] Error 2
> ```
> 
> I specified `CC=clang CXX=clang++` when I ran `cmake`.

Looks like setuptools picked the first compiler it found. I've updated the makefile to explicitly pass CC/CXX/LD that cmake detects and uses, to help it make consistent. I cannot check that myself, but hopefully this does the trick.

helios/pipeViewer/CMakeLists.txt
peter-d: It's removed.
I was no longer building argos_dumper, and that's been added in again now.
Does that need to be installed? Or is it used for in-tree tests only?

peter-d:
> Also, on the Mac, with `conda`, `pythonw` is required:

According to the setup tools docs this should now also be fixed, I'm using ```gui_script``` now vs ```console_script``` earlier in setup.py. Let me know if it works for you!

peter-d:
> $ mypy argos.py       
> model/element.py: error: Source file found twice under different module names: "pipe_view.model.element" and "model.element"
> model/element.py: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules for more info
> model/element.py: note: Common resolutions include: a) adding `__init__.py` somewhere, b) using `--explicit-package-bases` or adjusting MYPYPATH
> Found 1 error in 1 file (errors prevented further checking)

This error is because we run mypy in the same folder as where the pipe_view model lives. If I run it 1 folder up, I get rid of this error: 
```
mypy pipe_view/argos.py 
```

However that gives me 99 errors, both in this PR branch as on master. Might be an issue with the mypy ini file? It's almost all about wx and cython binary modules...

MichaelSchoenfelder-SiV:
Is https://github.com/sparcians/map/pull/406 causing problems with the the PipeViewer? It introduced new versions of Python and wxpython.

MichaelSchoenfelder-SiV:
@bdutro , the conda env in the CI should be installing boost 1.78.  Or were you also testing the non-conda build environment?

bdutro-sv:
@MichaelSchoenfelder-SiV I'm testing outside conda (we shouldn't need a conda env to build in a vanilla Ubuntu install).

bdutro:
I fixed the mypy errors and added some Argos linting jobs so we can catch this easily in the future.

peter-d:
> Is #406 causing problems with the the PipeViewer? It introduced new versions of Python and wxpython.

Not sure - I am not running on a conda based setup. But I am running it with python3.10 as well and it does work for me. 
I do get tiny tiny buttons and text in the gui on MacOs Ventura 3.1:
image

bdutro:
The next release of wx/wxPython will have better support for cross-platform DPI scaling. Right now your best bet is to just increase the font size if you're on Mac.

bdutro (approved):

knute-sifive:
If you're good with this @peter-d , I'll squash it today.

peter-d:
We should probably still update the docs somewhere to explain how to use the find_package(Sparta) in cmake.

Other than that: good to go for me! Thanks for all the contributions from the SiFive team to get this clean enough to merge!

knute-sifive:
Well, that's unfortunate.  I can't do this anymore:
```
cd map/sparta
mkdir release; cd release
cmake -DCMAKE_BUILD_TYPE=Release ..
```
The CMakeLists in sparta tree assume you're building at the `map` level:
```
CMake Error at simdb/test/HDF5Database/CMakeLists.txt:5 (include):
  include could not find requested file:

    /home/knutel/play/map/sparta/sparta/cmake/SimdbTestingMacros.cmake
```
Notice the double `sparta/sparta`.

This is important for us as our internal builds attempt to build ONLY the modeling framework and not the visualization tools.  I'll work on a clean fix.

peter-d:
Argh. I see. I was a bit confused by the makefiles structure. But knowing that, things make more sense…

I ran into 2 issues the way it was before:
1. Doing cmake at toplevel did not build or install everything as you would expect. 
2. If you don’t build it from the top map folder, the visualization tools make files don’t know where to find Sparta, as users could specify any folder to build. (I initially was building in a folder called debug and ran into issues in pipe view building, which had hardcoded release based paths)


Maybe the best way is to not have a toplevel cmake any more and only in Sparta and Helios folder.  And then give Helios cmake a parameter to find Sparta if needed?

knute-sifive:
> Maybe the best way is to not have a toplevel cmake any more and only in Sparta and Helios folder. And then give Helios cmake a parameter to find Sparta if needed?

I dunno... maybe that's the best thing to do: either build sparta or build Helios tools, not both.  But like you mentioned, the Argos tools might have issue with that.

I have a patch that tells the CMakeLists in sparta/test where to find these cmake files (no matter where Sparta is built from).

knute-sifive:
I think the installation should _always_ come from the `map` build/level though.

peter-d:
> I think the installation should _always_ come from the `map` build/level though.

So then there is at toplevel 2 folders:
- sparta
- helios

And there's a cmake flow in both that needs to be able to run stand-alone, but we need to enable a make install from the top?

peter-d:
> > And there's a cmake flow in both that needs to be able to run stand-alone, but we need to enable a make install from the top?
> 
> Really the `sparta` build can be a separate install, but I wasn't thinking that was a requirement. I guess it would be kinda dumb to only support install from the top level `map` environment (installing both Argos + sparta) and not just sparta. What if someone uses sparta and not Argos? And what if they can't get Argos to build? Frustrating!

I agree - it needs a clean enough separation to not force users to use both if they don't want to. (Sorry I updated my last comment still while you were typing your reply it seems...)

But then I propose: no toplevel cmake, and a separate one for sparta and helios folder. Then they are clearly decoupled, and users can install what they want.

knute-sifive:
> But then I propose: no toplevel cmake, and a separate one for sparta and helios folder. Then they are clearly decoupled, and users can install what they want.

Ya know, let's go for that divide.  That makes sense to me.

peter-d:
Shall I update the files accordingly? Might be useful if you push your patch for the tests, then I will take it from there.
(probably tomorrow - almost end of the workday already here)

knute-sifive:
Already pushed: https://github.com/sparcians/map/pull/397/commits/373f7f9e5375127656ef6dec7830089b5d0926e1

knute-sifive:
Thanks again -- no hurry on this.  I want to merge this in tomorrow (we're addressing some env issues here at SiFive and I want to fully test this in that environment).

peter-d:
Now separated out the builds. 

There is no longer a `CMakeLists.txt` at `map` level. Updated readme as well. The logic in the cmake is now the following:
1. You build sparta in `map/sparta/release` (or another subfolder of your choice in `map/sparta`)
2. Optionally: `make install` will install sparta in `CMAKE_INSTALL_PREFIX`
3. Build helios folder: `map/helios/release`(or another subfolder of your choice in `map/sparta`)
     - **NOTE**: This will by default look in system wide places for sparta (eg if you installed in `/usr/local` or in `-DSPARTA_SEARCH_DIR=<location of your choice>`, and if it fails to find that, look for sparta in the source tree, in that case always in `map/sparta/release`. Caveat here is that if you build in a folder not called release, then this breaks. (but with an error message to explain you need to set `SPARTA_SEARCH_DIR`)
5. Optionally: `make install` will install argos (and the python packages) in whatever python path is set up in your virtual/conda env.

peter-d:
Looks like CI wants there to be a central cmake entry point in the `map` folder - so CI breaks, but I don't know where to change this.

knute-sifive:
> Looks like CI wants there to be a central cmake entry point in the `map` folder - so CI breaks, but I don't know where to change this.

I can work on it.  Thanks!

knute-sifive:
Getting closer!  The ArgosDumper tool needs the `rt` library in Linux, but it's not part of the SPARTA::sparta library list.  Also, `simdb` isn't found for the transactiondb.so link.

knute-sifive:
Yahoo!!!  100% pass!

peter-d:
Great! I can confirm it still builds nicely on my end as well. One minor nitpick, I get (got it before as well I think, so it's not a side effect from the last few commits) a warning from setuptools on `pyrex_gdb=True`, and I see it's there in CI as well:
[`UserWarning: Unknown Extension options: 'pyrex_gdb' ](https://app.circleci.com/pipelines/github/sparcians/map/318/workflows/cdbd170d-356f-4557-a9e7-1c851f549295/jobs/247?invite=true#step-103-2669)

Other than that: looks good to go for me!
Thank you for dotting and crossing the final I'd and T's!

knute-sifive:
Gonna shoot for merging this PR this week.

peter-d:
Anything I can/should do to help?

knute-sifive:
Nope!  About the push the Big Green Button!  We needed to shut down auto-pulls and testing with our models here.

knute-sifive:
Actually, Peter, do you mind putting up your PR for the risc-v performance model again?

peter-d:
It’s there: https://github.com/riscv-software-src/riscv-perf-model/pull/42

sparta/simdb/test/CoreDatabase/CMakeLists.txt
```diff
@@ -1,8 +1,7 @@
 project(SIMDB_CoreDatabase_test)
 
 add_executable(SIMDB_CoreDatabase_test CoreDatabase_test.cpp)
-
-include(../TestingMacros.cmake)
+include(${SPARTA_CMAKE_MACRO_PATH}/SimdbTestingMacros.cmake)
```
timsnyder, line 4: Overall, this PR fixes quite a few things but I'm coming across this.  Putting `SimdbTestingMacros.cmake` in Sparta's `cmake/` subdirectory and defining `SPARTA_CMAKE_MACRO_PATH` in the toplevel `sparta/CmakeLists.txt` makes it so that simdb can't be built and regressed on it's own as a "cmake project".  @klingaard Was this intended?  Seems slightly odd if not broken to me.

PR #405, opened 2023-02-16
Move configurable initializations after RTN tree insertion

### LINKED ISSUE #404: Factory configurable order w.r.t. parameter initialization.
A subtle issue, but a pain if not fixed esp if users are using Factories to build tree hierarchies **and** they query the given TreeNode for parameter values to direct the topology.  Right now, the Parameter value will be the default value and not the latest from the command line, config file, or arch file.

When a user creates a `ResourceTreeNode` with a given factory, the RTN is placed into the tree _after_ that object's `Factory::onBuilding` is called.  Specifically, the `ParameterSet` for that RTN is created _before_ the RTN is placed in the tree.  Why this is a problem?  When the `ParameterSet` is created, each `Parameter` searches the virtual parameter tree for override values.  But the path the `Parameter` is using _has not been resolved_ since it's parent, the RTN has not been placed into the tree yet.  So... no match, no overrides.

The fix is to add the RTN to the tree first, _then create_ the `ParameterSet.`

knute-sifive:
closes #404

### DIFF
--- a/sparta/sparta/simulation/ResourceTreeNode.hpp
+++ b/sparta/sparta/simulation/ResourceTreeNode.hpp
@@ -79,11 +79,11 @@ namespace sparta
                     << getLocation() << " cannot be null";
             }
 
-            initConfigurables_();
-
             if(parent != nullptr){
                 parent->addChild(this); // Do not inherit parent state
             }
+
+            initConfigurables_();
         }
 
         //! \brief Alternate Constructor


davidsifive (approved):
LGTM

knute-sifive:
Once we have migrated our internal simulation tools to MAP master, this will be merged in.  We're almost done!

PR #408, opened 2023-02-20
moved WeightedContextCounter - fixes #407

### LINKED ISSUE #407: WeightedContextCounter: move from examples to sparta proper
Counter is defined in ```test/ContextCounter/WeightedContextCounter.hpp```

But it's used by https://github.com/riscv-software-src/riscv-perf-model (in ```Dispatch.cpp```) so it should probably be part of sparta proper instead of an example and test.

peter-d:


knute-sifive:
COMPELTELY forgot to do this.  So sorry.

knute-sifive (approved):

PR #410, opened 2023-02-28
test buffer works like std::list

shuzhih:


knute-sifive:
I have a fix in PR #397 that I'm working on for this regression fail.

knute-sifive (approved):

PR #415, opened 2023-03-27
Bump cmake version requirement to 3.19

peter-d:
- simpler boost detection on mac
- cmake clean up: target specific include paths
- boost 1.76
- boost 1.72 specific magic no longer needed
- Merged Peter's master branch with some cmake fixes on my end
- Moved reliance on include paths back to sparta.cmake file
- Made simdb include a private include
- PUBLIC includes for Sparta and SimDB to pass to regression
- Small cleanup; added missing SimDB testers target include
- cmake file to enable find_package(Sparta) in other projects
- make sure we include the testing macros
- also installing sparta/cache
- make sure core setup.py actually uses the interpreter found by find_package(PythonInterp)
- more modern way to find python3, hopefully also robust in conda/venv
- cleaner cmake for testing macros for simdb and sparta
- helios build now using FindSparta.cmake
- pipeviewer cannot use find_package(sparta), which only works once installed
- installing all py files and libs but module paths don't work yet after install
- simpler setup for cython build - working for transactiondb, rest todo
- cleaner HDF5 libs config for sparta
- Major rehaul of how argos ties into cmake
- Fixed static method calls w/o object
- Move __is_mac_os and __gen_message outside of the ShortcutHelp class
- Clean up flake8 errors
- make sure setuptools and cython use the right C/CXX/LD
- proper dependencies for setup.py
- gui script vs console script: should properly call pythonw
- build argos_dumper again
- _ vs __ otherwise cannot find the free helper functions in the class
- bumping cmake version to 3.19, needed to ensure HDF5::HDF5 target exists


peter-d:
Sorry this dragged in a whole bunch of old commits somehow - closing...

PR #416, opened 2023-03-27
bumping cmake version to 3.19, needed to ensure HDF5::HDF5 target exists

### LINKED ISSUE #407: WeightedContextCounter: move from examples to sparta proper

peter-d:
Fixes #407

### DIFF
--- a/helios/CMakeLists.txt
+++ b/helios/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 3.15)
+cmake_minimum_required (VERSION 3.19)
 
 project(helios)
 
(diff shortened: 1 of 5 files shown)


knute-sifive:
Did you mean to say it fixes #407 ?

knute-sifive:
I don't see issue with this -- we're using cmake 3.25 in house. Let me do a little testing just to make sure and I'll merge it in!  Thanks!

knute-sifive (approved):
LGTM

peter-d:
Just to double check: what's the policy here? Am I allowed to merge myself once approved?

knute-sifive:
> Just to double check: what's the policy here? Am I allowed to merge myself once approved?

Yep.  If I've given that power, you are free to use it.  😁

Issue #418, opened 2023-04-14
Failed to build map on Apple Silicon


furuame:
I followed steps of README to build a conda environment, but ran into the failure that CMake can't build map/sparta:

```
(sparta) m1:release zen$ CC=clang CXX=clang++ cmake .. -DCMAKE_BUILD_TYPE=Release
-- The CXX compiler identification is Clang 14.0.6
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - failed
-- Check for working CXX compiler: /Users/zen/miniconda3/envs/sparta/bin/clang++
-- Check for working CXX compiler: /Users/zen/miniconda3/envs/sparta/bin/clang++ - broken
CMake Error at /Users/zen/miniconda3/envs/sparta/share/cmake-3.26/Modules/CMakeTestCXXCompiler.cmake:60 (message):
  The C++ compiler

    "/Users/zen/miniconda3/envs/sparta/bin/clang++"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: /Users/zen/Sparta/map/sparta/release/CMakeFiles/CMakeScratch/TryCompile-VSm92n
    
    Run Build Command(s):/Users/zen/miniconda3/envs/sparta/bin/cmake -E env VERBOSE=1 /Users/zen/miniconda3/envs/sparta/bin/make -f Makefile cmTC_22bf8/fast && /Users/zen/miniconda3/envs/sparta/bin/make  -f CMakeFiles/cmTC_22bf8.dir/build.make CMakeFiles/cmTC_22bf8.dir/build
    make[1]: Entering directory '/Users/zen/Sparta/map/sparta/release/CMakeFiles/CMakeScratch/TryCompile-VSm92n'
    Building CXX object CMakeFiles/cmTC_22bf8.dir/testCXXCompiler.cxx.o
[...]

Issue #319, opened 2022-03-28
Are you supporting macOS on ARM?


timsnyder:
Just FYI, conda-forge is starting to build stuff for what they call `osx_arm64`.  I don't know if we're going to start having people with ARM based macbooks who need to develop stuff.

jackjchen:
It would help if I could build locally for osx-arm64, but I am not so much needing CI for this target platform. I tried to test https://github.com/sparcians/map/blob/master/scripts/create_conda_env.sh by hacking https://github.com/sparcians/map/blob/master/.ci_support/osx_64_.yaml with a few relevant values from https://github.com/conda-forge/conda-feedstock/blob/main/.ci_support/osx_arm64_python3.8.____cpython.yaml

It seems that the conda render does not expand the wildcard specs for python in the `build` section (`host` section is fine), and then the resulting environment.yml has the 2 incompatible merge specs for the conda-env create step:
   - python=3.8.\*=\*_cpython
   - python=3.8.13=h7e095e3_0_cpython

I was able to continue command line env creation by removing 1 of the python deps (and the no-spec texlive-core)

Does it seem this is an issue with the render from conda/conda for osx-arm64?

timsnyder-siv:
I'll take a look and see if I can reproduce the problem on my Mac.

klingaard:
Some folks in our Taiwan office have ARM-based Macs.  Not sure if they've tried the argos tools (which is what you really need the conda env for).  @2henwei

2henwei:
I did try building conda environment and encounter some issues, but I wasn't using `osx_arm64`. Did some dirty works to build the conda eventually, but `argos` just didn't work well (it could launch, but had some rendering issues).

knute-sifive:
Hey @timsnyder, is this issue good for closure?

jackjchen:
Issue looks solved regarding my earlier comment -- the scripts/create_conda_env.sh works fine for my local osx-arm64 dev environment after #419.

If anyone has interest, a separate issue would be for Linux-aarch64. There are problems with the create script with the conda solver not able to solve the rendered deps/versions and wxpython is not even found on conda-forge for aarch64.

PR #419, opened 2023-04-15
add osx_arm64 platform to conda-forge config

timsnyder:
Starting to address #418 and #319


conda-forge.yml
```diff
@@ -10,9 +10,12 @@ skip_render:
 provider:
   linux_64: circle
   osx_64: azure
+  osx_arm64: azure
   win: None
   linux_ppc64le: None
   linux_aarch64: None
+build_platform:
+  osx_arm64: osx_arm64
```
timsnyder, line 18: conda-forge has `build_platform.osx_arm64 = osx_64` by default and if you don't include this, you don't get the additional config when you rerender.

Still need to see if the rendered files will correctly build on an `osx_arm64` agent.  I suspect that they may fail spectacularly because native osx_arm64 doesn't seem to be fully supported by Azure Pipelines yet. ( see https://github.com/microsoft/azure-pipelines-agent/issues/3184)

timsnyder:
Looking at the [log for the first CI build](https://dev.azure.com/sparcians/map/_build/results?buildId=959&view=logs&j=ced5d8de-8227-5f3f-33a1-45cf1592c45a&t=1f4c3150-ddc9-52b5-8700-e07f78624442) of `osx_arm64`, it is definitely still cross-compiling.  I see:
```
CONDA_TOOLCHAIN_BUILD=x86_64-apple-darwin13.4.0
CONDA_TOOLCHAIN_HOST=arm64-apple-darwin20.0.0
```

I'll poke at this some more later.

timsnyder:
I think the azure pipelines agent that I pointed to earlier is only used for external runners that you run the agent on to hook them into your CI workflow.  It seems like M1 support on GitHub Actions (and also Azure Pipelines) is [coming later in 2023 (currently marked as Q4)](https://github.com/github/roadmap/issues/528).

knute-sifive (approved):
I'll go ahead and approve so when you're ready you can push.  Thanks for doing this Tim!

.ci_support/linux_64_.yaml
```diff
@@ -5,7 +5,7 @@ boost_cpp:
 c_compiler:
 - gcc
 c_compiler_version:
-- '11'
+- '12'
```
timsnyder, line 8: The latest rerender in e393be322fba59fd2f85a33831944e56ab94cd7f picked up new gcc and llvm.  If this is problematic, we can pin them back to versions 11 and 14 respectively.

timsnyder:
@knute-sifive looks like bumping to gcc 12 hits a compile error at https://app.circleci.com/pipelines/github/sparcians/map/389/workflows/383107d7-42ba-4f03-a697-a7e4b65a195c/jobs/295?invite=true#step-103-1815 

Should I pin back to gcc 11 for now in this PR and update to newer gcc and llvm in a follow-on or do you want me to just try to do the compiler update here also?

klingaard:
Whoa, full pass.  Nice!!  Ready to merge?

conda-forge.yml
timsnyder: I did some research yesterday and found that Github Actions supports arm64 on "Larger Instances" that Teams or Enterprise accounts can use but not on the free ones for open-source.  Azure Pipelines only supports them for self-hosted runners.

PR #426, opened 2023-05-10
Enhance MessageSource to control printing header

aka2en:
This would be useful when users want to customize the output of pevents.

### DIFF
--- a/sparta/sparta/pevents/PeventCollector.hpp
+++ b/sparta/sparta/pevents/PeventCollector.hpp
@@ -32,6 +32,7 @@ namespace pevents{
     class PeventCollector
         : public PairCollector<CollectedEntityType>, public PeventCollectorTreeNode
     {
+    protected:
         using PairCollector<CollectedEntityType>::getPEventLogVector;
         using PairCollector<CollectedEntityType>::turnOn_;
         using PairCollector<CollectedEntityType>::turnOff_;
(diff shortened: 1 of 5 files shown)


sparta/sparta/pevents/PeventCollector.hpp
```diff
@@ -180,14 +188,14 @@ namespace pevents{
          * \brief Override the generateCollectionString_() of the bases PairCollector,
          * In this method, we use the pair_cache owned by PairCache.
          */
-        virtual void generateCollectionString_() override final
+        virtual void generateCollectionString_() override
```
aka2en, line 191: Removing `final` as users may want to define customized string.

sparta/src/Destination.cpp
```diff
@@ -51,26 +51,29 @@ void DefaultFormatter::write(const sparta::log::Message& msg)
 {
     std::ios::fmtflags f = stream_.flags();
 
-    stream_ << '{';
+    if(msg.print_info)
```
aka2en, line 54: Users may want to output headerless message

knute-sifive: This kinda mimics the `sparta::log::RawFormatter`'s behavior...

knute-sifive (commented):
I _think_ see what you're doing here, but did you try this with the simulator?
```
--pevents pevents.log.raw ALL
```

aka2en:
`sparta::log::RawFormatter` can achieve what I want!

Issue #432, opened 2023-06-03
Clang default in README


dingiso:
CMake initially use GCC as compiler. Howerver, GCC was causing some false compilation errors. When I switched to Clang, Sparta compiled successfully. 

So I highly suggest setting clang as the default compiler, especially since it has already been included in the Conda environment.

The compile command in the README need to be updated accordingly.

```bash
CC=clang CXX=clang++ cmake .. -DCMAKE_BUILD_TYPE=Release
```

klingaard:
Sparta should compile cleanly with almost any compiler given.  What are the false errors you've observed?  Which version of `gcc` did you use?

dingiso:
Hi Knute,

These are some errors I got, I'm rethinking that some of them is not false. 

I'm using gcc 11.3.0

```bash
gcc version 11.3.0 (Ubuntu 11.3.0-1ubuntu1~22.04.1)
```

1.  Constructor in RegisterBits.

https://github.com/sparcians/map/blob/7b7f842b33e1e40e4cc7d3c73f9219b3172d4686/sparta/sparta/functional/RegisterBits.hpp#L79-L82

Using `local_storage_` to assign `local_data_` without constructing it first 

I fix it by adding local_storage_() between line 79 and 80, the compilation passed
 
2.  ss never be NULL in LexicalCast.hpp

https://github.com/sparcians/map/blob/60dd0303896730023289cdfffe80404bb23b557c/sparta/sparta/utils/LexicalCast.hpp#L78-L87

I fix it by deleting the line 85-87, it passed.

3.  fall through at switch in ConfigApplicators.hpp

https://github.com/sparcians/map/blob/60dd0303896730023289cdfffe80404bb23b557c/sparta/sparta/app/ConfigApplicators.hpp#L310-L316  

The last case and default fall through the switch without break.
But I think the first param in assert is 0 so it may be a false error 

I change the compiler to clang and it all gone.

knute-sifive:
1. Looks like a good fix
2. I have no idea what that cast does.  I did some research and I _think_ the intent was to check for a bad `stringstream`, but the code should look like this:  
```
    if(false == ss.fail()){
        return tmp;
    }
```
3. I think adding a `break` before line 314 will fix it

Do you mind making these patches and submitting a PR?

dingiso:
Hi Knute,

Great , I will making these patches and submitting a PR 

Thanks,
Dingisoul

klingaard:
Issues addressed.

PR #434, opened 2023-06-07
fix several errors (#432)

dingiso:
Hi Knute,

I fix these three errors and test on my machine, all looks good.

Thanks,
Dingisoul


knute-sifive (approved):
Appreciate the cleanup!

PR #435, opened 2023-07-17
Add debug-roi option support for pipeline collection

furuame:
Add `debug-roi` option to enable pipeline collection for multiple ROIs in a program. Users have to define ROI by themselves by sending notification of `sparta::roi::START` and `sparta::roi::STOP`. This PR hasn't added support for pevents and loggers. Enabling either options with `debug-roi` will raise the exception. They will be supported later.

With new option `debug-roi` enabled for pipeline collection, the pipeout files will be supplied with a integer suffix meaning the ROI index. Here is an example of `-z pipeout/` execution:
```
zen@m1: $ ls pipeout/
0_index.bin   1_index.bin   2_index.bin   0_clock.dat  0_display_format.dat  0_map.dat         1_clock.dat  1_display_format.dat  1_map.dat         2_clock.dat  2_display_format.dat  2_map.dat         simulation.info
0_record.bin  1_record.bin  2_record.bin  0_data.dat   0_location.dat        0_string_map.dat  1_data.dat   1_location.dat        1_string_map.dat  2_data.dat   2_location.dat        2_string_map.dat
```

~~Adding a PipelineNotifSrcTrigger to collect data of multiple ROIs. Instead of Triggers, it is driven by NotificationSource. Users will have to post notifications of `sparta::roi::START` and `sparta::roi::STOP` to tell pipeline collector when to start/stop collecting for an ROI.~~

~~As there can be many ROIs in a program, PipelineNotifSrcTrigger will generate pipeout dump "per ROI". ROI names are simply defined by putting the current number of ROI after the pipeout file path.~~

### DIFF
--- a/sparta/sparta/app/AppTriggers.hpp
+++ b/sparta/sparta/app/AppTriggers.hpp
@@ -139,7 +139,6 @@ private:
     const bool multiple_triggers_;
     sparta::Clock * clk_;
     sparta::RootTreeNode * root_;
-    bool triggered_ = false;
     uint32_t num_collections_ = 0;
 };
 
--- a/sparta/sparta/app/AppTriggers.hpp
+++ b/sparta/sparta/app/AppTriggers.hpp
@@ -32,15 +32,17 @@ public:
     PipelineTrigger(const std::string& pipeline_collection_path,
                     const std::set<std::string>& pipeline_enabled_node_names,
                     uint64_t pipeline_heartbeat,
+                    bool multiple_triggers,
                     sparta::Clock * clk, sparta::RootTreeNode * rtn) :
         pipeline_collection_path_(pipeline_collection_path),
         pipeline_enabled_node_names_(pipeline_enabled_node_names),
         pipeline_heartbeat_(pipeline_heartbeat),
+        multiple_triggers_(multiple_triggers),
         clk_(clk),
         root_(rtn)
     {
         pipeline_collector_.
-            reset(new sparta::collection::PipelineCollector(pipeline_collection_path_,
+            reset(new sparta::collection::PipelineCollector(multiple_triggers_ ? getCollectionPath_() : pipeline_collection_path_,
                                                             pipeline_heartbeat_,
                                                             clk_,
                                                             root_));
@@ -49,10 +51,33 @@ public:
 
     void go() override
     {
+        sparta_assert(!triggered_, "Why has pipeline trigger been triggered?");
         triggered_ = true;
         std::cout << "Pipeline collection started, output to files with prefix '"
                   << pipeline_collector_->getFilePath() << "'" << std::endl;
+        startCollection_();
 
+        if(multiple_triggers_) {
+            std::cout << "#" << num_collections_ << " pipeline collection started" << std::endl;
+        }
+    }
+
+    void stop() override
+    {
+        sparta_assert(triggered_, "Why stop an inactivated trigger?");
+        triggered_ = false;
+        stopCollection_();
+
+        if(multiple_triggers_) {
+            std::cout << "#" << num_collections_ << " pipeline collection ended" << std::endl;
+            ++num_collections_;
+            pipeline_collector_->reactivate(getCollectionPath_());
+        }
+    }
+
+private:
+    void startCollection_()
+    {
         if(pipeline_enabled_node_names_.empty()) {
             // Start collection at the root node
             pipeline_collector_->startCollection(root_);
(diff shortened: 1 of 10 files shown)


sparta/src/CommandLineSimulator.cpp
```diff
@@ -1130,7 +1139,7 @@ bool CommandLineSimulator::parse(int argc,
                 throw_report_deprecated = true;
                 ++i;
 
-            }else if (o.string_key == "pipeline-collection") {
+            }else if (o.string_key == "pipeline-collection" or o.string_key == "roi-pipeline-collection") {
```
furuame, line 1142: Both options share the if condition, so I can guard that only one of them is enabled.

knute-sifive:
This is awesome!  But this being specific to pipeline collection seems limiting. I can see requests for pevent, logging, etc where an ROI is handy to enable.  The issue is -- what defines an ROI?  Obviously that's up to the framework user.   As an example, for core-based simulations, that might be instructions (see [CSEM_INSTRUCTIONS](https://github.com/sparcians/map/blob/e3d5358dd2a30fa42ffde5edb350d2fc41aef070/sparta/sparta/app/Simulation.hpp#L74)) or for a MSS simulation, this might be transactions, etc.

Thinking about this some more, this is actually related to the `--debug-on` command line option, which defines just 1 ROI (or at least the start of one).  If you're looking to create _multiple_ ROIs, it might be best served by the realized simulator instead of the framework...

furuame:
> This is awesome! But this being specific to pipeline collection seems limiting. I can see requests for pevent, logging, etc where an ROI is handy to enable. The issue is -- what defines an ROI? Obviously that's up to the framework user. As an example, for core-based simulations, that might be instructions (see [CSEM_INSTRUCTIONS](https://github.com/sparcians/map/blob/e3d5358dd2a30fa42ffde5edb350d2fc41aef070/sparta/sparta/app/Simulation.hpp#L74)) or for a MSS simulation, this might be transactions, etc.

That's a good point ... I think a general way to define an ROI would be: interface. Let users tell framework 1) when to start? and 2) when to end? So that's why I added the definitions of START/END in `RegionOfInterest.hpp` and adopted NotficationSource. If we need multiple ROIs support for pevent and logging, we could just add the corresponding implementations (driven by NotificationSource) in sparta. If users have already utilized the interface, they could just enjoy the enhancements then.

knute-sifive:
Chatted with @furuame on the side...

Zhen and I bounced around some ideas on how to make this more generic, i.e. supporting more than just pipeline collection.  We settled on a new command line option: `--debug-on-roi` that falls into the pattern of `--debug-on <cycle>` and `--debug-on-inst <inst>`.  (Perhaps `--debug-on-roi` can take an argument?  Maybe the ROI count?  I dunno).



To add/connect the `debug-on-roi` trigger, add a new variable to CommandLineSimulator class called something like `trigger_on_roi_` and set it true when the command line option is provided.  Also create your `PipelineTrigger` object.

Then, to [this line](https://github.com/sparcians/map/blob/e3d5358dd2a30fa42ffde5edb350d2fc41aef070/sparta/src/CommandLineSimulator.cpp#L2118), add a check for that variable.  In the body of that `if` statement, add the created `PipelineTrigger` to the `debug_trigger_` object.

Then, in the `switch` statement below,  instead of using `setTriggerStartAbsolute`, create a new method in the debug trigger that uses a notification source and pass in ROI.

sparta/src/CommandLineSimulator.cpp
```diff
@@ -1361,6 +1373,18 @@ bool CommandLineSimulator::parse(int argc,
                         << o.value.at(0) << "\"";
                 }
                 ++i;
+            } else if(o.string_key == "debug-on-roi") {
+                if(delayed_start)
+                {
+                    std::cerr << MULTI_INSTRUCTION_TRIGGER_ERROR_MSG << std::endl;
+                }
```
knute-sifive, line 1380: Wondering if these checks should be asserts.  Seems the code continues running if this were true.

sparta/src/CommandLineSimulator.cpp
```diff
@@ -362,6 +367,14 @@ CommandLineSimulator::CommandLineSimulator(const std::string& usage,
...
          "Note: Any directories in this path must already exist.\n",
          "Enable pipline collection to files with names prefixed with OUTPATH") // Brief
+        ("roi-pipeline-collection,Z",
+         named_value<std::vector<std::string>>("OUTPUTPATH", 1, 1)->multitoken(),
+         "Run pipeline collection for ROIs on this simulation, and dump the output files to OUTPUTPATH. "
+         "OUTPUTPATH can be a prefix such as myfiles_ for the pipeline files and may be a "
+         "directory\n"
+         "Example: \"--pipeline-collection data/test1_\"\n"
+         "Note: Any directories in this path must already exist.\n",
+         "Enable pipline collection to files with names prefixed with OUTPATH") // Brief
```
knute-sifive, line 377: What's this for?

knute-sifive:
@furuame , @bdutro-sv fixed the compilation issue.  Merge in master with your branch and let's see what happens!

furuame:
@bdutro-sv Thank you!!!

furuame:
Since we haven't supported `debug-roi` for pevents and loggers, I have tested both options in my use case. The sparta correctly captured them and showed the error messages.

sparta/src/CommandLineSimulator.cpp
```diff
@@ -2122,7 +2144,7 @@ void CommandLineSimulator::populateSimulation_(Simulation* sim)
                 debug_trigger_->addTriggeredObject(pevent_trigger_.get());
             }
 
-            // Pipeline trigger
+            // Pipeline trigger.
```
furuame, line 2147: Not intentionally ... but I'm lazy to revert it 😂

sparta/src/CommandLineSimulator.cpp
```diff
@@ -2153,6 +2175,17 @@ void CommandLineSimulator::populateSimulation_(Simulation* sim)
...
                 break;
+            case SimulationConfiguration::TriggerSource::TRIGGER_ON_ROI:
+                {
+                    try{
+                        debug_trigger_->setTriggerNotificationDriven(sim->getRoot(), roi::NOTIFICATION_SRC_NAME);
+                    }catch(SpartaException& ex) {
+                        std::cerr << "\nTo use debug-roi option, users have to register notification source for ROI" << std::endl;
+                        std::cerr << "\n\n" SPARTA_CMDLINE_COLOR_ERROR "Rethrowing..." SPARTA_CMDLINE_COLOR_NORMAL << std::endl;
+                        throw;
+                    }
```
furuame, line 2186: @knute-sifive does it work for you?

sparta/sparta/app/AppTriggers.hpp
```diff
@@ -75,37 +103,45 @@ class PipelineTrigger : public trigger::Triggerable
...
+            return pipeline_collection_path_ + '_' + std::to_string(num_collections_) + '_';
+        }
+    }
 
     std::unique_ptr<collection::PipelineCollector> pipeline_collector_;
+    const std::string pipeline_collection_path_;
     const std::set<std::string> pipeline_enabled_node_names_;
+    const uint64_t pipeline_heartbeat_;
+    const bool multiple_triggers_;
+    sparta::Clock * clk_;
```
knute-sifive, line 140: Should initialize to `nullptr`

sparta/src/CommandLineSimulator.cpp
```diff
@@ -1300,8 +1304,8 @@ bool CommandLineSimulator::parse(int argc,
                 }
                 delayed_start = true;
 
-                if(sim_config_.trigger_on_value != static_cast<uint64_t>(SimulationConfiguration::TriggerSource::TRIGGER_ON_NONE)) {
-                    throw SpartaException("Cannot use both --debug-on and --debug-on-icount simultaneously");
+                if(sim_config_.trigger_on_type != SimulationConfiguration::TriggerSource::TRIGGER_ON_NONE) {
+                    throw SpartaException("Cannot use either --debug-on, --debug-on-icount or --debug-on-roi simultaneously");
```
knute-sifive, line 1308: I think the best language here is:

"Cannot use --debug-on, --debug-on-icount, and --debug-on-roi simultaneously"

sparta/src/Simulation.cpp
```diff
@@ -480,6 +480,12 @@ void Simulation::configure(const int argc,
                                                sim_config_->warnings_file));
     }
 
+    // FIXME: Support debug-roi for loggers
+    if(sim_config_->trigger_on_type == SimulationConfiguration::TriggerSource::TRIGGER_ON_ROI &&
+       !sim_config_->getTaps().empty()) {
+        throw SpartaException("debug-roi hasn't been supported for loggers");
```
knute-sifive, line 486: Can you change the error message to:
"Logging ennoblement is currently not supported with debug-roi.  Use --debug or --debug-on-icount"

sparta/src/CommandLineSimulator.cpp
knute-sifive: Yep!  Thanks!

knute-sifive (approved):
Approved, but some small tweaks are requested.

sparta/sparta/app/AppTriggers.hpp
furuame: Done

sparta/src/CommandLineSimulator.cpp
furuame: Done

sparta/src/Simulation.cpp
furuame: And done

PR #442, opened 2023-09-20
ALFLayout Generator

knute-sifive:
New python script that replaces the old, outdated perl generator.

This new generator consumes a `location.dat` file from a generated pipeout and, when coupled with a custom python script, can be used to explicitly setup ALF files used by the Argos tools.

See `pydoc ALFLayout` for more information as well as the example in example.CoreModel


sparta/example/CoreModel/gen_layouts.py
```diff
@@ -0,0 +1,87 @@
+import sys,os
+
+sys.path.append('../../../../helios/pipeViewer/scripts')
```
kathlenehurt-sifive, line 3: Is it possible to specify an absolute path here with a variable instead of the relative path?

kathlenehurt-sifive:
I'd recommend adding a main function in `gen_layouts.py` incase you ever want to add command line parameters.

sparta/example/CoreModel/gen_layouts.py
```diff
@@ -0,0 +1,87 @@
...
+        ./sparta_core_example -i1 -z pipeout/
+    ''')
+    exit(255)
+
+try:
+    os.mkdir('layouts')
+except:
+    pass
+
+num_cycles = 55
```
kathlenehurt-sifive, line 19: Variables intended to be constant should be named in all caps.

kathlenehurt-sifive (approved):
Neat!

sparta/example/CoreModel/gen_layouts.py
knute-sifive: Yeah, I'll change it to build the abs path.

sparta/example/CoreModel/gen_layouts.py
knute-sifive: Done

sparta/example/CoreModel/gen_layouts.py
knute-sifive: Doine

knute-sifive:
> I'd recommend adding a main function in `gen_layouts.py` incase you ever want to add command line parameters.

Not a bad idea.  Wish I were a better python programmer.  Not sure what I'd do with a `main`.  😄

PR #451, opened 2023-10-26
Implement PriorityQueue

### LINKED ISSUE #450: Implement a PriorityQueue
stdc++ has a handy `priority_queue` class, but it's available in C++20, which we haven't moved to (yet).  But that class has some restrictions and a sparta supported PriorityQueue might be beneficial.

This is especially handy for the [Olympia performance model](https://github.com/riscv-software-src/riscv-perf-model)

CC: @h0lyalg0rithm

klingaard:
Also addresses some issues with FastList's iterator class.

closes #450


sparta/sparta/resources/PriorityQueue.hpp
```diff
@@ -0,0 +1,186 @@
...
+         * \param data Reference to the data object forced to the
+         *             front of the queue
+         *
+         * Push the data item to the front of the queue, bypassing the
+         * internal sorting algorithm.  This will mess up the sorting
+         * -- be warned.
+         */
+        void forceFront(const DataT & data) {
+            priority_items_.emplace_front(data);
+        }
```
furuame, line 161: Interesting. Wondering the usage? If an entry can be pushed like this way, a `refresh` method to sort up entries might sound useful for users.

klingaard: I thought about a `resort` call, but I'd rather someone request it.  :smile: 

This method is useful in situations like coherency models, where multiple requests from many sources might come in and be sorted properly.  But then, at the last minute, an I/O device really needs attention due to a possible buffer overflow (I'm making stuff up :wink: ) and must be highest priority even though it actually can be considered lower.

sparta/test/PriorityQueue/PriorityQueue_test.cpp
```diff
@@ -0,0 +1,210 @@
...
+    sparta::PriorityQueue<uint32_t> pqueue;
+
+    for(auto i : {1,3,2,5,6,4,8,7}) {
+        pqueue.insert(i);
+    }
+
+    EXPECT_EQUAL(pqueue.size(), 8);
+
+    EXPECT_EQUAL(pqueue.top(), 1);
+    pqueue.pop();
```
phattso, line 23: Minor but to be consistent with the other lines of code:

```
pqueue.pop(); // 1
```

sparta/test/PriorityQueue/PriorityQueue_test.cpp
```diff
@@ -0,0 +1,210 @@
...
+    EXPECT_EQUAL(pqueue.top(), -7);
+
+    dyn_sorter.toggleSmallerFirst();
+
+    pqueue.insert(11);
+    EXPECT_EQUAL(pqueue.top(), 11);
+
+    // for(auto i : pqueue) {
+    //     std::cout << i << std::endl;
+    // }
```
phattso, line 145: Remove?

phattso (approved):

sparta/test/PriorityQueue/PriorityQueue_test.cpp
klingaard: added

sparta/test/PriorityQueue/PriorityQueue_test.cpp
klingaard: Useful debug in a tester.  I'll let it ride :car:

Issue #311, opened 2022-03-08
In a report YAML, cannot refer to a new SI in a subsequent expression


knute-sifive:
In a Report YAML, a user cannot define a new statistic definition/instance and the refer to it in a subsequent expression:
```
content:
   cpu.core0.ctrs:
        counter_foo: foo
        "foo + cycles": foo_with_cycles
```
In the above example, `foo` is not known to the Report (not a valid `TreeNode`).  This is because the `ExpressionGrammar` is only given the search node hierarchy and not a list of newly formed `StatisticInstances` created by the Report class.

knute-sifive:
@TopChangwatchaiSiFive first noticed this.  

Some caveats (beyond the general issue): once `foo` is defined, it can be referenced in _subsequent expressions_.  If an expression refers to `foo` _before_ it's defined, that will be an error (and hopefully expected to be one).

PR #456, opened 2023-11-17
Fix for issue #311: Make an StatisticInstance report reference

klingaard:
Before this fix, users could not refer to a newly defined StatisticInstance in a report in subsequent expressions:
```
content:
   cpu.core0.ctrs:
        counter_foo: foo
        "foo + cycles"         : foo_with_cycles
        "foo_with_cycles + 10" : foo_with_cycles_plus_10    # Previously this was an error: foo_with_cycles is unknown
```


sparta/src/Report.cpp
```diff
@@ -445,24 +445,24 @@ class ReportFileParserYAML
...
+                                Report* const r = report_map_.at(scope.uid);
+                                statistics::expression::Expression expr(assoc_key, n, r->getStatistics());
                                 StatisticInstance si(std::move(expr));
                                 si.setContext(n);
                                 std::string full_name = value;
                                 auto& captures = scope.second;
-                                Report* const r = report_map_.at(scope.uid);
                                 if(getSubstituteForStatName(full_name, n, captures)){
                                     r->add(si, full_name);
                                 }
```
bdutro, line 456: This code appears multiple times (407-414, 436-443, 460-468). It might be worth factoring out into a function.

sparta/src/Report.cpp
```diff
@@ -1159,6 +1159,157 @@ class ReportFileParserYAML
...
+
+    // Track a new stat with helpful exception wrapping
+    addField_(name, si);
+
+    if(name != ""){ stat_names_.insert(name); }
+
+    addSubStatistics_(&si);
+
+    return Report::StatAdder(*this);
+}
```
bdutro, line 1179: `Report::add` might be worth templatizing. It looks like several of the overrides share the same code.

sparta/src/StatisticInstance.cpp
```diff
@@ -0,0 +1,770 @@
...
+                                         const InstrumentationNode::class_t cls,
+                                         const std::vector<std::pair<std::string, std::string>> & metadata)
+    {
+        provided_location_ = location;
+        provided_description_ = description;
+        provided_expr_string_ = expression_str;
+        provided_value_semantic_ = value_semantic;
+        provided_visibility_ = visibility;
+        provided_class_ = cls;
+        provided_metadata_ = metadata;
```
bdutro, line 220: These could be member initializers.

sparta/src/StatisticInstance.cpp
```diff
@@ -0,0 +1,770 @@
...
+    }
+
+    double StatisticInstance::getValue() const {
+        if (direct_lookup_si_value_ != nullptr) {
+            return computeValue_();
+        }
+
+        if(end_tick_ < start_tick_){
+            throw ReversedStatisticRange("Range is reversed. End < start");
+        }
```
bdutro, line 345: This seems like something that could be checked when `end_tick_` and `start_tick_` are assigned, unless those change frequently and the `getValue` is called infrequently.

sparta/src/StatisticInstance.cpp
```diff
@@ -0,0 +1,770 @@
...
+
+        if(start_tick_ > getScheduler_()->getElapsedTicks()){
```
bdutro, line 347: Use `SPARTA_EXPECT_FALSE` here

sparta/src/StatisticInstance.cpp
```diff
@@ -0,0 +1,770 @@
...
+            throw FutureStatisticRange("Range starts in the future at ") << start_tick_;
+        }
+
+        double value;
+        if(end_tick_ == Scheduler::INDEFINITE){
+            // Compute Value
+            value = computeValue_();
+        }
+
+        else if(end_tick_ > getScheduler_()->getElapsedTicks()){
```
bdutro, line 357: Use `SPARTA_EXPECT_FALSE` here

sparta/sparta/statistics/StatisticInstance.hpp
```diff
@@ -1168,41 +560,41 @@ namespace sparta
...
          */
-        const CounterBase* ctr_;
+        const CounterBase* ctr_ = nullptr;
 
         /*!
          * \brief Parameter reference from which this statistic instance will
          * compute its value (nullptr if there is no Parameter to reference)
          */
-        const ParameterBase* par_;
+        const ParameterBase* par_ = nullptr;
```
bdutro, line 575: It looks like these three members are never used at the same time. Either `sdef_` is used, or `ctr_` is used, or `par_` is used. If these don't share a common base class with virtual methods, it might make sense to add a templatized wrapper class that can handle this instead of having if-else-if-else-if everywhere.

sparta/src/Expression.cpp
```diff
@@ -14,16 +14,25 @@ Expression::Expression(const std::string& expression,
                        TreeNode* context)
 {
     std::vector<const TreeNode*> already_used;
-    parse_(expression, context, already_used);
+    const std::vector<stat_pair_t> report_si;
```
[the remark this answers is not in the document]
klingaard: :thinking: True, it's only a placeholder.  Will save a little time.  Thanks!

sparta/src/Expression.cpp
```diff
@@ -37,10 +46,12 @@ Expression::~Expression() {
 
 void Expression::parse_(const std::string& expression,
                         TreeNode* context,
-                        std::vector<const TreeNode*>& already_used) {
+                        std::vector<const TreeNode*>& already_used,
+                        const std::vector<stat_pair_t> &report_si)
+{
```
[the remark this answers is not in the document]
klingaard: True, but won't that just duplicate the body in multiple places?  Maybe add defaults to the parameters...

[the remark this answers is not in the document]
klingaard: But that won't with a non-const reference, so nvmd

sparta/src/Expression.cpp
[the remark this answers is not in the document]
klingaard: Done

sparta/src/Expression.cpp
```diff
@@ -14,16 +14,25 @@ Expression::Expression(const std::string& expression,
...
+    const std::vector<stat_pair_t> report_si;
+    parse_(expression, context, already_used, report_si);
 }
 
 Expression::Expression(const std::string& expression,
                        TreeNode* context,
                        std::vector<const TreeNode*>& already_used)
 {
-    parse_(expression, context, already_used);
+    const std::vector<stat_pair_t> report_si;
```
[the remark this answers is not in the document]
klingaard: Done

sparta/src/Expression.cpp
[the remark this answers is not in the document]
bdutro: My thinking was you could move the placeholders into the overloads instead of having to duplicate them everywhere you call `parse_`.

klingaard:
closes #311

sparta/src/ExpressionGrammar.cpp
```diff
@@ -220,6 +224,12 @@ class lazy_gen_var_
                 n = calculator->getNode();
             }
         }
+        auto it = std::find_if(report_si_.begin(), report_si_.end(), [a1] (const auto & si_pair) {
+                                                                         return (si_pair.first == a1);
+                                                                     });
```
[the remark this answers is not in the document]
klingaard: You mean like this?
```
        if(auto it = std::find_if(report_si_.begin(),
                                  report_si_.end(),
                                  [&a1] (const auto & si_pair) {
                                      return (si_pair.first == a1);
                                  }); it != report_si_.end())
```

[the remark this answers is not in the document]
bdutro: Yep!

sparta/src/Report.cpp
klingaard: Yeah, I noticed that too.  One of those -- "more work than I wanted to invest" but since I'm under the hood...

sparta/src/ExpressionGrammar.cpp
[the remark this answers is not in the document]
klingaard: Done

sparta/src/Report.cpp
klingaard: I 100% agree with you.  However, I think I'll save that for another day.  This is code I originally didn't write/own, and I just did some pretty big surgery on it.

sparta/src/Report.cpp
klingaard: Done

sparta/src/StatisticInstance.cpp
klingaard: Done

sparta/src/StatisticInstance.cpp
klingaard: Done

sparta/src/StatisticInstance.cpp
klingaard: Done

sparta/sparta/statistics/StatisticInstance.hpp
klingaard: Yeah, I can see the original implementation could have used some better design.  I think this is also beyond the scope of this PR.  I'll append this to the existing issue I opened.

sparta/src/Report.cpp
klingaard: New Issue opened

sparta/src/Expression.cpp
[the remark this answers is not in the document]
klingaard: I think I'll leave this alone for now.  Every re-implementation I come up with seems to be roughly the same outcome,.,

klingaard:
@bdutro -- while I look into why conda-forge is crapping out on MacOS, do you have any other inputs to this PR?  I appreciate what you found!  If not, can you approve it?

bdutro (approved):

PR #460, opened 2023-12-19
Fix building step commands

lhtin:
Hi,

When I follow the building steps, I encounter some errors like bellow, this PR fix these errors.

1. Step 3 already cd to `map` dir, so step 7 `cd map/sparta` should change to `cd sparta`. Step 8 has the same problem.
2. Step 6 already active the `sparta` conda env, so step 7 `conda activate sparta` is no needed. Step 8 has the same problem.
3. In step 7 `cd map/sparta; mkdir release; cd release` , failure of a previous command does not prevent a subsequent command from running when using `;`  to run multiple command. Changing to `&&` is more friendly since we want all commands success. Step 8 has the same problem.

Best,
Lehua

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,6 @@ installed and would like to build everything (not just sparta).
    * `make`
    * `cmake --install . --prefix $CONDA_PREFIX`
 1. To build Helios/Argos transaction viewer:
-   * `conda activate sparta`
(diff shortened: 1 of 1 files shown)


klingaard:
Appreciate your update to the docs!

klingaard (approved):

PR #467, opened 2024-01-14
Changes for Sparta @ C++23

acrucker:
Two changes need to be accounted for:

1. C++20's DR 2237, which forbids writing constructors as templates.
2. std::is_pod<T> is deprecated, and needs to be replaced using std::is_trival<T> && std::is_standard_layout<T>

### DIFF
--- a/sparta/simdb/include/simdb/schema/ColumnMetaStructs.hpp
+++ b/sparta/simdb/include/simdb/schema/ColumnMetaStructs.hpp
@@ -11,6 +11,7 @@
 #include "simdb/schema/ColumnTypedefs.hpp"
 #include "simdb/schema/DatabaseTypedefs.hpp"
 #include "simdb/Errors.hpp"
+#include "simdb/utils/CompatUtils.hpp"
 
 #include <string>
 #include <type_traits>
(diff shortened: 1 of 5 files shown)


sparta/simdb/include/simdb/schema/ColumnMetaStructs.hpp
```diff
@@ -23,14 +23,20 @@ namespace simdb {
 template <typename ColumnT, typename Enable = void>
 struct column_info;
 
+template<typename T>
+struct is_pod {
+  static constexpr bool value = std::is_trivial<T>::value
+                             && std::is_standard_layout<T>::value;
+};
+
```
klingaard, line 31: Wonder if this could be added to sparta/sparta/utils/MetaStructs.hpp reused in the multiple places you've changed.

One concern I have with this definition here -- it's not in a namespace, which means if someone did something silly like `using namespace std;` and then included this file, the compiler might see multiple definitions.

klingaard (commented):
Thank you for the fixes!  I have one comment I'd like you to consider.

sparta/simdb/include/simdb/schema/ColumnMetaStructs.hpp
acrucker: I have moved the struct to a new file in the simdb include directory. This is because it appeared that simdb files were being included in both simdb and sparta, but I didn't see the sparta path being passed as an include for the simdb compilation. The struct is now located at "simdb::utils::is_pod"

klingaard (approved):
Thank for you the fixes!  LGTM.

PR #468, opened 2024-01-18
Scoreboard View Fix

AaronGChan:
Scoreboard view is hardcoded to look for scoreboards along `core.*` only. I hit this bug when working on development work for [Olympia](https://github.com/riscv-software-src/riscv-perf-model), where our unit tests initialize a simulation without a core node, thus resulting in the [sparta assert](https://github.com/sparcians/map/blob/4626e0dbbee6ebd07bf035e4ed9b85ac959abbb3/sparta/src/Scoreboard.cpp#L317) triggering. The proposed fix is to check if the `cpu_node` is null, then use the root node to search for scoreboards. An additional test for a cpu not named "core" is added to verify these changes.

### DIFF
--- a/sparta/src/Scoreboard.cpp
+++ b/sparta/src/Scoreboard.cpp
@@ -309,14 +309,8 @@ namespace sparta
 
         // Try to find the master scoreboard, if it's available (has
         // been created by the Sparta framework)
-
-        // Go as high as the CPU node in this Tree.  If we go higher,
-        // we could bind to a Scoreboard in another CPU!  That'd be
-        // bad.
-
-        // Search from the root
-        auto cpu_node = parent->getRoot();
-        sparta_assert(cpu_node != nullptr, "Could not find the core nodes in this simulation");
+        
+        // Search for scoreboards from parent
 
         std::function<Scoreboard*(sparta::TreeNode *)> findScoreboard =
             [&] (sparta::TreeNode * node) -> Scoreboard * {
@@ -350,7 +344,7 @@ namespace sparta
                 return scoreboard;
             };
 
-        Scoreboard * master_sb = findScoreboard(cpu_node);
+        Scoreboard * master_sb = findScoreboard(parent);
 
         // Gotta be more than 0
         sparta_assert(master_sb != nullptr,
--- a/sparta/src/Scoreboard.cpp
+++ b/sparta/src/Scoreboard.cpp
@@ -314,6 +314,10 @@ namespace sparta
         // we could bind to a Scoreboard in another CPU!  That'd be
         // bad.
         auto cpu_node = parent->findAncestorByName("core*");
+        // if a core node is not available, search from the root
+        if(cpu_node == nullptr){
+            cpu_node = parent->getRoot();
+        }
         sparta_assert(cpu_node != nullptr, "Could not find the core nodes in this simulation");
 
         std::function<Scoreboard*(sparta::TreeNode *)> findScoreboard =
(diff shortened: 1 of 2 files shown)


sparta/src/Scoreboard.cpp
```diff
@@ -314,6 +314,10 @@ namespace sparta
         // we could bind to a Scoreboard in another CPU!  That'd be
         // bad.
         auto cpu_node = parent->findAncestorByName("core*");
+        // if a core node is not available, search from the root
+        if(cpu_node == nullptr){
+            cpu_node = parent->getRoot();
+        }
```
klingaard, line 320: Hey Aaron, thanks for the fix.  I think you can remove the query to find the `findAncestorByName` call completely.  Just start with the `parent` node in the search (i.e. remove the concept of `cpu_node` completely).  See if that works.

AaronGChan: If we are in a multicore system, wouldn't this run into the issue of multiple scoreboards being found?

klingaard: Technically the current way has this problem.  I think just removing the initial path check will work and possibly keep the matches to the core the view is contained in.

sparta/src/Scoreboard.cpp
```diff
@@ -313,7 +313,9 @@ namespace sparta
         // Go as high as the CPU node in this Tree.  If we go higher,
         // we could bind to a Scoreboard in another CPU!  That'd be
         // bad.
-        auto cpu_node = parent->findAncestorByName("core*");
+
+        // Search from the root
+        auto cpu_node = parent->getRoot();
```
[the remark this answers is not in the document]
AaronGChan: So currently in Olympia, when we do a scoreboard view, the `parent` that we pass in is the current node, i.e if we're calling it from [ExecutePipe](https://github.com/riscv-software-src/riscv-perf-model/blob/ce29ae25e3e026db1c4e36f48068e68efef2d091/core/ExecutePipe.cpp#L65), the `parent` would be an `alu0` node for example. So the change above would then require us to pass from the function caller side the appropriate parent node that we know contains the scoreboard.

I see the benefit of this change, in that it makes this function more generalized and lets the coder define it in the case of a multicore system, but just wanted to double check that this was the intended change.

sparta/test/Scoreboard/Scoreboard_test.cpp
```diff
@@ -152,7 +152,7 @@ class ExeUnit : public sparta::Unit
         my_scoreboard_view_.reset
             (new sparta::ScoreboardView(getContainer()->getName(), // ALU0, ALU1, LSU, FPU, etc
                                         sb_unit_type_,             // integer, fp, vector
-                                        getContainer()));          // Used to find the Scoreboard
+                                        getContainer()->getRoot()));          // Used to find the Scoreboard
```
klingaard, line 155: _Technically_ you can call `getRoot` here, but this is a tester, no harm.  Also, did it not work with just `getContainer`?

klingaard (approved):
@furuame any comments?

sparta/test/Scoreboard/Scoreboard_test.cpp
AaronGChan: `getContainer` didn't work. I think it was too far down the tree, so the scoreboard was above.

furuame (approved):
> @furuame any comments?

Oops ... good catch! LGTM!

PR #474, opened 2024-02-27
Remove unused member ev_producing_event_(SkeletonPipeline)

lhtin:
The class member ev_producing_event_ of Producer is unused, the 1 cycle delay is configured in Consumer `consumer_in_port_` member like below. Removing it to avoid misunderstanding.

```c++
class Consumer : public sparta::Unit
{
    // Consumer's InPort to get data
    sparta::DataInPort<uint32_t> consumer_in_port_{&unit_port_set_, "consumer_in_port", 1};
}
```

### DIFF
--- a/sparta/example/SkeletonPipeline/src/Producer.hpp
+++ b/sparta/example/SkeletonPipeline/src/Producer.hpp
@@ -61,10 +61,6 @@ private:
     // Producer's producer handler
     void produceData_();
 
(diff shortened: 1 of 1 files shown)


klingaard (approved):
Yeah, I think I had an intention of using this event to show how you can delay internal work in the producer.  But we have other examples of that elsewhere.

Thanks for the fix.

PR #483, opened 2024-03-15
Fix syntax error warning causing error on build for some compiler versions (older GCC versions) 

adrenaholicgeek:


klingaard (approved):
Simple enough.  Thought I got all of those.  :)

PR #488, opened 2024-04-05
Fix the script to make sure user is using conda-forge channel.

adrenaholicgeek:
This resolves the further dependency issues.


adrenaholicgeek:
The Issue was:

klingaard:
@timsnyder ,do you mind approving this?

klingaard (approved):
Looks like a great change to me.  @timsnyder I don't think you'll have issues with this.

Thanks for the fix!

PR #489, opened 2024-04-08
Add an implementation of a true LRU cache replacement policy

bdutro:


sparta/test/cache/simple_cache/main.cpp
```diff
@@ -15,7 +15,7 @@ dl1_(32,  // 32KB in size
      LINE_SIZE,  // line size
      LINE_SIZE,  // stride
      sparta::cache::LineData(LINE_SIZE), // line_size
-     sparta::cache::TreePLRUReplacement(8) );   // num_ways
+     sparta::cache::LRUReplacement(8) );   // num_ways
```
klingaard, line 18: Does this mean the TreePLRU is no longer tested?

bdutro: The `l2_` and `btb_b_` instances in that test still use TreePLRU. I figured we could leave both in place.

klingaard (approved):

PR #493, opened 2024-04-23
Define custom copy/move constructors/operators for SharedData and GlobalEvent

bdutro:
I ran into a segfault when I tried to create an `std::vector` of `SharedData`. There were two underlying causes for this:

1. When `GlobalEvent` is moved/copied, it doesn't update `ev_handler_lifetime_` to point to the new instance of `event_handler_`. If the original `GlobalEvent` was later destroyed, the copy would segfault when it tried to fire its handler.
2. When `SharedData` is moved/copied, its `ev_update_` receives a copy of the handler in the original `SharedData`. If the original was later destroyed, the copy would segfault when it tried to perform its state update.

This PR defines copy/move constructors and assignment operators for both classes so that everything gets updated properly and the copies can continue to function after the originals are destroyed.

### DIFF
--- a/sparta/sparta/events/GlobalEvent.hpp
+++ b/sparta/sparta/events/GlobalEvent.hpp
@@ -106,6 +106,43 @@ namespace sparta
             ev_sched_ptr_(clk->getScheduler()->getGlobalPhasedPayloadEventPtr<sched_phase_T>())
         {}
 
+        GlobalEvent(const GlobalEvent& rhs) :
+            local_clk_(rhs.local_clk_),
+            event_handler_(rhs.event_handler_),
+            ev_handler_lifetime_(&event_handler_),
+            ev_sched_ptr_(rhs.ev_sched_ptr_)
+        {
+        }
+
+        GlobalEvent(GlobalEvent&& rhs) :
+            local_clk_(rhs.local_clk_),
+            event_handler_(std::move(rhs.event_handler_)),
+            ev_handler_lifetime_(&event_handler_),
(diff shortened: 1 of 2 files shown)


sparta/sparta/events/GlobalEvent.hpp
klingaard, line 1: With the exception of the move constructor, can these not be `= default`?

sparta/sparta/resources/SharedData.hpp
```diff
@@ -65,6 +70,44 @@ namespace sparta
             writePS(std::forward<U>(init_val));
         }
 
+        SharedData(const SharedData& rhs) :
+            ev_update_(rhs.ev_update_),
```
klingaard, line 74: Why copy the ev_update_ if it's going to be reset?

klingaard:
Good catch.  This has been plaguing me for quite some time!

sparta/sparta/resources/SharedData.hpp
bdutro: We still need the other fields inside `ev_update_` to get copied.

sparta/sparta/events/GlobalEvent.hpp
bdutro: No, we have to explicitly initialize `ev_handler_lifetime_` to point to the `ev_handler_` inside the new instance in all of these cases. Otherwise, it will point to `rhs.ev_handler_`.

klingaard: Ah, I see the difference.  Thanks!

klingaard (approved):

PR #497, opened 2024-04-30
Fix reporting bug where ContextCounters have their per-context counts printed twice

bdutro:
I ran into a bug where the per-context counts from a ContextCounter would get printed twice in a report:

```
              my_counter = 882885
              my_counter_context0 = 221216
              my_counter_context1 = 220455
              my_counter_context2 = 220124
              my_counter_context3 = 221090
              my_counter.context0 = 221216
              my_counter.context1 = 220455
              my_counter.context2 = 220124
              my_counter.context3 = 221090

```

This PR fixes the report generator so that only the second set (with the `.` separator) is included in the report.

### DIFF
--- a/sparta/src/Report.cpp
+++ b/sparta/src/Report.cpp
@@ -1150,7 +1150,7 @@ private:
     std::string filename_; //!< For recalling errors
 }; // class ReportFileParserYAML
 
-Report::StatAdder Report::add(const StatisticInstance& si, const std::string& name) {
+Report::StatAdder Report::add(const StatisticInstance& si, const std::string& name, const bool recurse) {
     if(name != "" && stat_names_.find(name) != stat_names_.end()){
         throw SpartaException("There is already a statistic instance in this Report (")
             << getName() << ") named \"" << name << "\" pointing to "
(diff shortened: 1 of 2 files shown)


sparta/src/Report.cpp
```diff
@@ -1416,7 +1420,7 @@ void Report::recursAddSubtree_(const TreeNode* n,
             //for(auto& s : stats_){
             //    std::cerr << "  Stat " << s.second->stringize() << " " << s.second->getLocation() << std::endl;
             //}
-            add(n, child_stat_prefix + n->getName());
+            add(n, child_stat_prefix + n->getName(), false);
```
bdutro, line 1423: `recursAddSubtree_` is already recursively walking the entire tree, so there's no need for `add` to also recursively add its children.

klingaard: Is there a location where the recurse _needs_ to be true?

bdutro: Thanks for this, it made me realize what was really going on:

If you look [here](https://github.com/sparcians/map/pull/497/files#diff-d8d78a9a996d832adee697448c329df4c5cffd5aa39b3cff52afddc051d9705aR1392-R1419) you'll see that we're calling `add()` where the first argument is a `const TreeNode*`. However, we only have a non-const `TreeNode*` overload for `add` [here](https://github.com/sparcians/map/pull/497/files#diff-8e1255d7e3412b851f8a5453e15a77e697369a4bbccab9e5e86e2ce435b67c13L438-R438). Because `StatisticInstance` has a non-explicit constructor that takes a `const TreeNode*`, we actually end up calling the `add` overload for `StatisticInstance&&`.

If you look at the `TreeNode*` overload for `add` [here](https://github.com/sparcians/map/pull/497/files#diff-d8d78a9a996d832adee697448c329df4c5cffd5aa39b3cff52afddc051d9705aL1228-R1244), you'll see that there isn't any recursion happening at all. So I think the original intent was to use this version, but the const-ness was wrong.

My latest commit undoes the extra `recurse` parameter and just changes the `add` signature to take `const TreeNode*`.

bdutro: I tried changing all the single-argument constructors in StatisticInstance.hpp to be `explicit` and everything seems to work fine (regress passes, everything compiles properly). Do you want me to add that to this PR as well to prevent some of this weirdness in the future?

klingaard: Yeah, please do!

klingaard (approved):
Approval pending on the explicit construction...

sparta/src/Report.cpp
bdutro: Done

PR #498, opened 2024-04-30
Fix compilation errors with Clang v16+. Also includes fixes for compiling with -std=c++23.

bdutro:
This PR fixes compilation errors I observed trying to compile with clang v16 and v17, as well as some errors from bumping the C++ standard to C++23.

### DIFF
--- a/sparta/sparta/pipeViewer/Outputter.hpp
+++ b/sparta/sparta/pipeViewer/Outputter.hpp
@@ -187,7 +187,7 @@ namespace sparta::pipeViewer
                     if(!dat.stringVector[i].empty()){
                         // We check if we have seen this exact pair, field and value before or not.
                         if(const auto& [val, str] = std::tie(dat.valueVector[i].first, dat.stringVector[i]);
-                           stringMap.try_emplace(std::make_tuple(dat.pairId, i, val), str).second) {
+                           stringMap.emplace(std::piecewise_construct, std::forward_as_tuple(dat.pairId, i, val), std::forward_as_tuple(str)).second) {
                             // We add this mapping into out String Map file which we will
                             // use when reading back from the database.
                             string_file_ << dat.pairId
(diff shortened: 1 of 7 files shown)


sparta/sparta/pipeViewer/Outputter.hpp
```diff
@@ -187,7 +187,7 @@ namespace sparta::pipeViewer
                     if(!dat.stringVector[i].empty()){
                         // We check if we have seen this exact pair, field and value before or not.
                         if(const auto& [val, str] = std::tie(dat.valueVector[i].first, dat.stringVector[i]);
-                           stringMap.try_emplace(std::make_tuple(dat.pairId, i, val), str).second) {
+                           stringMap.emplace(std::piecewise_construct, std::forward_as_tuple(dat.pairId, i, val), std::forward_as_tuple(str)).second) {
```
bdutro, line 190: Workaround for a Clang bug: https://github.com/llvm/llvm-project/issues/61415

This is essentially what `try_emplace` does under the hood, so performance shouldn't be affected.

sparta/sparta/resources/Buffer.hpp
```diff
@@ -95,7 +95,7 @@ namespace sparta
          */
         struct DataPointer {
         private:
-            typename std::aligned_storage<sizeof(value_type), alignof(value_type)>::type object_memory_;
+            alignas(value_type) std::byte object_memory_[sizeof(value_type)];
```
bdutro, line 98: `std::aligned_storage` was deprecated in C++23: https://stackoverflow.com/questions/71828288/why-is-stdaligned-storage-to-be-deprecated-in-c23-and-what-to-use-instead

sparta/sparta/utils/Utils.hpp
```diff
@@ -27,56 +27,56 @@
 /*!
  * \brief Custom literal for uint64
  */
-constexpr inline uint64_t operator "" _u64(unsigned long long i) {
+constexpr inline uint64_t operator ""_u64(unsigned long long i) {
```
bdutro, line 30: Clang throws a compiler error if there is a space between the `""` and the `_`

sparta/src/StatisticsArchives.cpp
```diff
@@ -196,6 +196,15 @@ void RootArchiveNode::saveTo(const std::string & dir)
...
 
+ArchiveNode::ArchiveNode(const std::string & name) :
+    name_(name)
+{
+}
+
+ArchiveNode::~ArchiveNode()
+{
+}
+
```
bdutro, line 207: Moving these to the `.cpp` file helps avoid some pitfalls when you use an incomplete type with a smart pointer.

sparta/sparta/utils/SpartaSharedPointerAllocator.hpp
klingaard, line 1: Nice cleanup!

klingaard (approved):

PR #500, opened 2024-05-02
Fix NameError: name 'ScheduleLineElement' is not defined error in Argos

bdutro:
Fixes the following error that can occur when advancing Argos quickly:

```
NameError: name 'ScheduleLineElement' is not defined
```

### DIFF
--- a/helios/pipeViewer/pipe_view/model/schedule_element.py
+++ b/helios/pipeViewer/pipe_view/model/schedule_element.py
@@ -16,7 +16,9 @@ from .element import (Element,
                       PropertyValue,
                       ValidatedPropertyDict)
 from . import element_propsvalid as valid
-from .element_value import Element_Value, FakeElementValue
+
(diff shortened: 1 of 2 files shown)


helios/pipeViewer/pipe_view/model/schedule_element.py
```diff
@@ -387,6 +389,8 @@ def GetQueryFrame(self, period: int) -> Tuple[int, int]:
     def DetectCollision(self,
                         pt: Union[Tuple[int, int], wx.Point],
                         pair: Element_Value) -> FakeElementValue:
+        from .element_value import FakeElementValue
+
```
bdutro, line 393: `FakeElementValue` is only actually used in this function. By importing it here, we avoid a circular import dependency with `element_value.py`.

klingaard (approved):
Cool -- I came across this error every now and then, but found it to be a "stupid" on my end.

Issue #507, opened 2024-07-17
Allow for optional content in a report file


klingaard:
Would be really handy to use the same report definition file for multiple configurations that might or might not have the same stats.  Today the report definition causes the framework to error:
```
Unable to parse expression: "cpu.icache.stats.il1_cache_misses" within context: top in report file "report.yaml" for the following reason: 
SPARTA was unable to find a tree node that matched this name.
```
Here's what the report.yaml looks like:
```
name: "Summary ICache Report"
content:
  cpu.icache.stats.il1_cache_misses: ICache Misses
```
This works for architecture A which has the icache at the `cpu` level, but not architecture B which has it maybe inside `fetch`.

Suggested "fix":
```
name: "Summary ICache Report"
optional_content:
  # Works for arch A
  cpu.icache.stats.il1_cache_misses: ICache Misses
optional_content:
  # Works for arch B
  cpu.fetch.icache.stats.il1_cache_misses: ICache Misses
```

klingaard:
Implementation example:
```
name: "Summary Performance Report"
style:
  decimal_places: 2
content:
  cycles : Total cycles consumed
simple-arch-content:
  cpu.core*.rob.stats.total_number_retired : Total Instructions Retired
other-arch-content:
  cpu.core*.rob.stats.ipc : IPC
```
The report mechanism looks for the keyword `<arch name>-arch-content` where `<arch name>` is the value supplied on the command line via `--arch <arch name>` flag

PR #508, opened 2024-07-23
Provide a way to append arch-specific stats to report definitions

colby-nyce:
Provides a fix for this issue:
https://github.com/sparcians/map/issues/507

### DIFF
--- a/sparta/src/Report.cpp
+++ b/sparta/src/Report.cpp
@@ -117,6 +117,14 @@ class ReportFileParserYAML
          */
         std::stack<bool> in_content_stack_;
 
+        /*!
+         * \brief Are we accepting the stats inside the current content block?
+         * An example of when we do not accept stats is when we are parsing
+         * a content block for an arch that does not match the --arch at the
+         * command line.
+         */
+        bool skip_content_leaves_ = false;
+
         // Did we find an 'ignore' block?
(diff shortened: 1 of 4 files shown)


sparta/src/Report.cpp
klingaard, line 1: Wonder if there's value in adding `arch-content` as a constexpr string (keyword) like `content`: `KEY_ARCH_CONTENT`.  Of course, you still need to perform the `find` operation as expected

sparta/src/Report.cpp
```diff
@@ -731,6 +750,52 @@ class ReportFileParserYAML
...
+                        }
+
+                        if (!sim) {
+                            throw SpartaException("Could not get the app::Simulation to parse key: ") << key;
+                        }
+
+                        if (auto sim_config = sim->getSimulationConfiguration()) {
+                            skip_content_leaves_ = true;
+                            bool dash_arch_given = false;
+                            for (const auto &kvp : sim_config->getRunMetadata()) {
```
klingaard, line 769: Might be handy to have sim config to enable a search of a key.

klingaard:
I don't know what's going on with azure.

klingaard:
All regressions passed locally (by hand).  I'll merge this in for now

PR #513, opened 2024-08-11
[ERROR] BasicCacheSet const

orion160:


klingaard:
Thanks, Isaac, for doing this. I actually fixed this in PR https://github.com/sparcians/map/pull/510 which I just merged. Can you rebase and see if there is still an issue?

orion160:
It works!

PR #515, opened 2024-08-29
Enable dynamic-sized RegisterBits

furuame:
Hello Knute,

In some vector processors, we may need registers longer than 512-bit. Was `std::array` used for considering C++ performance? It might work if I just upsize it to 1024-bit, but I think there could be larger ones in the future. Use `std::vector` instead could be once and for all.

Let me know your thoughts! I tried to make `64` as template parameter by default -- But Registerbits seems to be used heavily as return value type, and class template argument deduction can’t work for that case.

### DIFF
--- a/sparta/sparta/functional/Register.hpp
+++ b/sparta/sparta/functional/Register.hpp
@@ -1332,10 +1332,10 @@ protected:
 private:
     RegisterBits computeWriteMask_(const Definition *def) const
     {
-        // Non-power-of-2 size may cause illegal access,
-        // an exception will be thrown in the constructor.
         if(!isPowerOf2(def->bytes)) {
-            return RegisterBits(nullptr);
+            throw SpartaException("Register \"")
+                << getName() << "\" size in bytes must be a power of 2 larger than 0, is "
+                << def->bytes;
         }
 
         const auto mask_size = def->bytes;
--- a/sparta/sparta/functional/Register.hpp
+++ b/sparta/sparta/functional/Register.hpp
@@ -1332,6 +1332,12 @@ protected:
 private:
     RegisterBits computeWriteMask_(const Definition *def) const
     {
+        // Non-power-of-2 size may cause illegal access,
+        // an exception will be thrown in the constructor.
+        if(!isPowerOf2(def->bytes)) {
+            return RegisterBits(nullptr);
+        }
+
         const auto mask_size = def->bytes;
         RegisterBits write_mask(mask_size);
         RegisterBits partial_mask(mask_size);
--- a/sparta/sparta/functional/RegisterBits.hpp
+++ b/sparta/sparta/functional/RegisterBits.hpp
@@ -16,8 +16,8 @@ namespace sparta
      * \class RegisterBits
      *
      * This class is used in conjuntion with sparta::RegisterBase to
-     * quickly write masked registers of sizes between 1 and 512
-     * bytes.  This class replaces the use of BitArray.
+     * quickly write masked registers.  This class replaces the use
+     * of BitArray.
      *
      * The class works by assuming register data is handed to it via a
      * char array.  The class will "view" into this data until it's
@@ -63,9 +63,16 @@ namespace sparta
 
         // Copy the remote register data locally.
         void convert_() {
-            if(nullptr == local_data_) {
-                local_data_ = local_storage_.data();
-                ::memset(local_data_, 0, local_storage_.size());
+            if(nullptr == local_data_)
+            {
+                if(num_bytes_ <= local_storage_.size()) {
+                    local_data_ = local_storage_.data();
+                }
+                else {
+                    local_storage_alt_.reset(new uint8_t[num_bytes_]);
+                    local_data_ = local_storage_alt_.get();
+                }
+                ::memset(local_data_, 0, num_bytes_);
                 ::memcpy(local_data_, remote_data_, num_bytes_);
                 remote_data_ = local_data_;
             }
@@ -82,9 +89,12 @@ namespace sparta
             remote_data_(local_data_),
             num_bytes_(num_bytes)
         {
-            ::memset(local_data_, 0, local_storage_.size());
-            sparta_assert(num_bytes <= local_storage_.size(),
-                          "RegisterBits size is locked to 64 bytes. num_bytes requested: " << num_bytes);
+            if(num_bytes > local_storage_.size()) {
+                local_storage_alt_.reset(new uint8_t[num_bytes]);
+                local_data_ = local_storage_alt_.get();
+                remote_data_ = local_data_;
+            }
+            ::memset(local_data_, 0, num_bytes_);
         }
 
         /**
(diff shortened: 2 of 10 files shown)


klingaard:
Hey Zen!  Yes, `std::array` was used as the base for the class mostly because the backend data was inlined with the class (better locality).  But you can use a `std::vector` if you want as long as performance says sharp.

BTW, there' s a test: https://github.com/sparcians/map/blame/master/sparta/test/Register/reg_bit_test.cpp that you can build by hand to test it.  I forgot to hook it up to regression.

furuame:
> BTW, there' s a test: https://github.com/sparcians/map/blame/master/sparta/test/Register/reg_bit_test.cpp that you can build by hand to test it. I forgot to hook it up to regression.

Will do!

> Yes, std::array was used as the base for the class mostly because the backend data was inlined with the class (better locality). But you can use a std::vector if you want as long as performance says sharp.

std::vector version was down ~2% simulation performance on generating a 50M trace, that was a shame. I came up with another way to keep std::array, but dynamically allocate memory when the size > 64B ... testing it ...

If my trick works, that might be a good idea to shrink the std::array data to 8B as generally registers are 64-bit.

knute-mips:
Look forward to your trick.  :)

furuame:
Done! The new version was measured 6.8% faster than the master build on generating a 10M+50M instruction trace 🚀

sparta/sparta/functional/RegisterBits.hpp
```diff
@@ -697,9 +725,10 @@ namespace sparta
 
     private:
 
-        std::array<uint8_t, 64> local_storage_; //!< Local storage
-        uint8_t               * local_data_  = nullptr; //!< Points to null if using remote data
-        const uint8_t         * remote_data_ = nullptr; //!< Remove data; points to local_data_ if no remote
-        const uint64_t          num_bytes_ = 0; //!< Number of bytse
+        std::array<uint8_t, 8>   local_storage_; //!< Local storage
+        std::unique_ptr<uint8_t> local_storage_alt_; //!< Alternative local storage when register size > 64B
```
klingaard, line 729: I think you meant `std::unique_ptr<uint8_t[]>` here.

klingaard (changes requested):
I think there's a memory leak here.

sparta/sparta/functional/RegisterBits.hpp
furuame: Oops sorry -- My “C” brain thought I could use that as a uint8_t pointer!

klingaard:
I'll take a look at the regressions to see why it's failing.  It's another irritant, I think.

furuame:
> I'll take a look at the regressions to see why it's failing. It's another irritant, I think.

Huh it’s like on me. Weird -- I passed it on my Apple Silicon box.

```
 58/101 Test  #58: Register_test ....................Subprocess aborted***Exception:   0.10 sec
````

I remember I saw a different CI failure previously though. I’ll also take a look at this one tomorrow.

**Comment by klingaard** [comment]
I ran `valgrind --vgdb=full --vgdb-error=0  ./Register_test` and it found a few issues with RegisterBits.  Might start there.

**Comment by furuame** [comment]
Thank you! Will do!

**Comment by furuame** [comment]
> I ran `valgrind --vgdb=full --vgdb-error=0 ./Register_test` and it found a few issues with RegisterBits. Might start there.

Done! That one was tricky ... Valgrind rocks!!

Regressions still fail but that seems a Cython compilation issue on Argos ... @klingaard would you help take a look at it?

File: sparta/sparta/functional/Register.hpp [hunk 2]
```diff
@@ -1332,6 +1332,12 @@ class RegisterBase : public TreeNode
 private:
     RegisterBits computeWriteMask_(const Definition *def) const
     {
+        // Non-power-of-2 size may cause illegal access,
+        // an exception will be thrown in the constructor.
+        if(!isPowerOf2(def->bytes)) {
+            return RegisterBits(nullptr);
+        }
```
furuame, line 1339: Would it be better toI just throw a SpartaException or keep this as it is?

furuame: The write mask is initialized by Class initializer list, but the legal register size is later checked by the constructor ... 

However an odd size would cause illegal access here.

klingaard: I think you should throw.

klingaard:
> Regressions still fail but that seems a Cython compilation issue on Argos

Sigh... will take a look

klingaard:
I'm going to remove helios/pipeview from the regression build.  It's broken and we're re-writing it, so I don't care as much

sparta/sparta/functional/RegisterBits.hpp
```diff
@@ -150,10 +158,22 @@ namespace sparta
...
         RegisterBits(const RegisterBits & orig) :
-            local_storage_(orig.local_storage_),
-            local_data_(orig.local_data_ == nullptr ? nullptr : local_storage_.data()),
-            remote_data_(orig.local_data_ == orig.remote_data_ ? local_data_ : orig.remote_data_)
-        {}
+            num_bytes_(orig.num_bytes_)
+        {
+            if(num_bytes_ <= local_storage_.size()) {
+                local_storage_ = orig.local_storage_;
+                local_data_ = orig.local_data_ == nullptr ? nullptr : local_storage_.data();
```
klingaard, line 165: When does this happen?

klingaard: Curious, when can `num_bytes_` be <= local storage size and the original still have its local_data_ pointer set to `nullptr`?

klingaard (approved):

sparta/sparta/functional/Register.hpp
furuame: Agree, that’s much better

furuame:
I merged this into the `map_v2` branch.

sparta/sparta/functional/RegisterBits.hpp
knute-mips: But if it has > 0 bytes, will it ever be `nullptr`?

furuame: Yes, it could be if it is initialized by a `const uint8` pointer. In that case, only `remote_data_` pointer is initialized. `local_data_` is `nullptr` until value is changed.

knute-mips:
I haven't documented the release process fully, but in a nutshell (I will documented this officially 😄 ):

- Merged changes into `master`
- Merge changes from `master` into `map_v2` branch
- Tag `map_v2` branch with the next tag
- [Make a release](https://github.com/sparcians/map/releases)

furuame:
> I haven't documented the release process fully, but in a nutshell (I will documented this officially 😄 ):
> 
> * Merged changes into `master`
> * Merge changes from `master` into `map_v2` branch
> * Tag `map_v2` branch with the next tag
> * [Make a release](https://github.com/sparcians/map/releases)

I am glad I do it right so far!

Would we need a process to release a tag? I haven’t done it and filed a discussion here: #519

PR #522, opened 2024-09-13
Create registers from JSON definition

colby-nyce:


sparta/scripts/GenRegisterJSON.py
```diff
@@ -0,0 +1,498 @@
...
+                "name":          name,
+                "num":           num,
+                "desc":          self.get_int_reg_desc(num),
+                "size":          self.reg_size,
+                "aliases":       alias,
+                "fields":        fields,
+                "initial_value": 0,
+                "enabled":       True}))
+
+        #Include PC, Athena expects the PC to always be 64-bits
```
kathlenemagnus-mips, line 243: Can you delete lines 243-265? They don't need to be checked into Sparta.

colby-nyce: Done (including removing resv_valid)

sparta/sparta/functional/RegisterDefnsJSON.hpp
```diff
@@ -0,0 +1,235 @@
...
+        // Add a definition that indicates the end of the array
+        register_defns_.push_back(RegisterBase::DEFINITION_END);
+    }
+
+    RegisterBase::Definition* getAllDefns() {
+        return register_defns_.data();
+    }
+
+private:
+    void parse(const std::string& register_defns_json_filename)
```
kathlenemagnus-mips, line 37: Rename method to `parse_`.

sparta/sparta/functional/RegisterDefnsJSON.hpp
```diff
@@ -0,0 +1,235 @@
...
+        std::stringstream ss;
+        ss << ifs.rdbuf();
+        std::string json_str = ss.str();
+
+        // Parse the JSON string
+        rapidjson::Document document;
+        document.Parse(json_str.c_str());
+
+        for (auto& item : document.GetArray()) {
+            if (item.HasMember("enabled") && !item["enabled"].GetBool()) {
```
kathlenemagnus-mips, line 50: I think this may break backwards compatibility. @furuame can you confirm? From the functional model's POV, I believe there is a difference between "this register is invalid" and "this register is valid but not enabled for this arch".

sparta/sparta/functional/RegisterDefnsJSON.hpp
```diff
@@ -0,0 +1,235 @@
...
+            const char *name = cached_strings_.back().raw();
+
+            const RegisterBase::group_num_type group_num = item["group_num"].GetInt();
+            auto iter = group_idx_map_.find(group_num);
+            if (iter == group_idx_map_.end()) {
+                group_idx_map_[group_num] = 0;
+            }
+
+            RegisterBase::group_idx_type group_idx = group_idx_map_[group_num]++;
+            cached_strings_.emplace_back(item["group_name"].GetString());
```
kathlenemagnus-mips, line 66: Should the `group_name` be optional as well since `group_num` is?

kathlenemagnus-mips (approved):
LGTM. I have one concern about backwards compatibility, but I'm ok with fixing it in a later PR if necessary.

sparta/scripts/GenRegisterJSON.py
```diff
@@ -0,0 +1,463 @@
...
+"""Helper script for regenerating register definition JSON files.
+"""
+
+from enum import IntEnum
+import json
+import math
+import sys
+import pdb
+
+from RV64_CSR import CSR64_DEFS
```
kathlenemagnus-mips, line 11: These libraries are missing from the PR.

sparta/scripts/GenRegisterJSON.py
```diff
@@ -0,0 +1,463 @@
...
+            name = self.get_fp_reg_name(num)
+            alias = FP_ALIASES[name] if (name in FP_ALIASES.keys()) else []
+            self.reg_defs.append(self.__CreateRegDict({
+                "name":          name,
+                "num":           num,
+                "desc":          self.get_fp_reg_desc(num),
+                "size":          8,
+                "aliases":       alias,
+                "fields":        fields,
+                "initial_value": 0,
```
kathlenemagnus-mips, line 275: Does this need to be changed to `"x0"`?

sparta/scripts/GenRegisterJSON.py
```diff
@@ -0,0 +1,463 @@
...
+
+        for k, v in CSR_DEFS.items():
+            self.reg_defs.append(self.__CreateRegDict({
+                "name":          v[0],
+                "num":           k,
+                "desc":          v[1],
+                "size":          8,
+                "aliases":       ["csr"+str(k)],
+                "fields":        v[2],
+                "initial_value": v[3],
```
kathlenemagnus-mips, line 291: Same concern as above, need to make sure the initial values are hex strings.

sparta/sparta/functional/RegisterDefnsJSON.hpp
colbynyce-mips: Not sure I understand - I am adding "group_name" and "group_num" to every register json. It was my understanding that neither would be optional.

sparta/scripts/GenRegisterJSON.py
colbynyce-mips: In the __CreateRegDict method, I remove the "initial_value" key from the dict if it is zero, else convert to hex.

sparta/sparta/functional/RegisterDefnsJSON.hpp
furuame: Yes, I can confirm. This should be fine. We handle that scenario in JSON level. The `enabled` field is used to be overridden by the other arch JSON to turn that register off. The disabled register is never created.

PR #527, opened 2024-09-26
sparta::Buffer Enhancements

### LINKED ISSUE #464: `Sparta::Buffer::erase(const iter&)` should return updated iter reference
### Discussed in https://github.com/sparcians/map/discussions/459

<div type='discussions-op-text'>

<sup>Originally posted by **timsnyder** December 18, 2023</sup>
In https://github.com/sparcians/map/blob/e876f588ec53eab68c24e4f70adf47c3c7e901c2/sparta/sparta/resources/Buffer.hpp#L663-L683

The return types of `Sparta::Buffer::erase(const iter&)`  are `void`.  Based on the STL semantics for `erase()`, I would expect them to return a copy of the iterator adjusted for the erasure.  Does `Sparta::Buffer::erase()` invalidate the iterator?

For example, with a `std::vector`, I would expect to need to handle iterator invalidation something like:
```
#include <vector>


int main() {
    std::vector<int> junk = {1,3,5,7,9};

    for(auto itr = junk.begin(); itr != junk.end();) {
        if (*itr == 5) {
            itr = junk.erase(itr);
        } else {
            itr++;
        }
    }
}
```

Do I need to worry about the same type of thing with `Sparta::Buffer` or is the iterator somehow not invalidated by `Sparta::Buffer::erase()`?</div>

bdutro:
Makes `sparta::Buffer::erase()` and makes `sparta::Buffer::reverse_iterator` behave more like their STL counterparts.

Note that there is a quirk in the behavior of `reverse_iterator`:

```
for(int i = 0; i < 10; ++i)
{
    buf.push_back(i);
}
auto it = buf.rbegin();
std::cout << *it; // Prints out 9
buf.pop_back();
std::cout << *it; // Prints out 8
```

Because `sparta::Buffer::end()` is a constant and `sparta::Buffer::rbegin()` points to `end()`, it will always be valid after an `erase` or `pop_back` as long as the buffer is not empty. This makes using `pop_back` while looping over reverse iterators unintuitive. However, the following pattern works as expected:

```
auto it = buf.rbegin();
while(it != buf.rend())
{
    it = buf.erase(it);
}
```

Fixes #464


klingaard (approved):
Awesome testing.  Thank you!

PR #536, opened 2024-10-25
Add debugging support for DAG

furuame:
It would make modelers' lives easier to debug Event ordering by just enabling scheduler debug logger `-l scheduler debug 1`.

I implicitly turn on the early cycle detection when the debug logger is enabled. I think that would be useful since that seems never be enabled 🤔 There is a method to turn it on but you would need to re-compile the software during debugging. Let me know your thoughts.


furuame:
I just saw the test failures. I should keep the method to keep backward compatibility.

klingaard (approved):
LGTM

PR #537, opened 2024-10-25
Add knob to turn off auto precedence for Events

furuame:
There is already a method for InPorts and OutPorts to turn off auto precedence in Sparta Unit. However, it would also be good to have a similar method but only effective on Events.

### DIFF
--- a/sparta/sparta/simulation/Unit.hpp
+++ b/sparta/sparta/simulation/Unit.hpp
@@ -125,6 +125,12 @@ namespace sparta {
 
             for(auto & event_node : unit_event_set_.getEvents(sparta::SchedulingPhase::Tick))
             {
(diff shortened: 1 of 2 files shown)


klingaard (approved):
Good idea.

sparta/sparta/simulation/Unit.hpp
```diff
@@ -125,6 +125,12 @@ namespace sparta {
 
             for(auto & event_node : unit_event_set_.getEvents(sparta::SchedulingPhase::Tick))
             {
+                // This EventNode does not participate auto precedence
+                // between InPorts and OutPorts.
```
furuame, line 129:
```suggestion
                // This event does not participate in auto precedence.
```

PR #543, opened 2024-11-22
Add support for turning off signal handler that prints backtrace

VictorYing:
Enable modelers to disable Sparta's signal handlers, since it conflicts with other framework like SST. No change to default behavior.


VictorYing:
Thanks for the feedback. It let us to find an existing variable `SimulatorConfiguration::signal_mode` that seemed to exist for this purpose but have no effect previously, so we simply hooked it up in `Simulation::configure()` as you suggested.

klingaard (approved):
Looks good.  Thanks for the update.  I've asked Tim in the other PR to update MacOS CI.

klingaard:
Hey @VictorYing, can you pull from the latest master of sparta into your fork?  The CI issue has been resolved.

PR #546, opened 2024-12-12
Do not allow dmiWrite to touch x0

colby-nyce:


sparta/sparta/functional/Register.hpp
```diff
@@ -1081,7 +1081,9 @@ class RegisterBase : public TreeNode
     template <typename T>
     void dmiWrite(T val, index_type idx = 0)
     {
-        dmiWrite_(&val, sizeof(val), sizeof(val) * idx);
+        if (getID() || getGroupNum()) {
```
kathlenemagnus, line 1084: I don't like the idea of including RISC-V behavior in Sparta. Could we add a bool class member called `writable_` that we can set at construction time?

kathlenemagnus (approved):

PR #548, opened 2025-01-06
Fix -arch-content bug in report yaml parser

colby-nyce:


klingaard (approved):
Thanks for the quick fix!

PR #550, opened 2025-01-28
Map v3

colby-nyce:


colby-nyce:
Note to all: this is not a formal PR and should be ignored for now. I'm using it as a self-review and to make sure I don't lose weeks' worth of work.

colby-nyce:
Closing, moving to my own fork

PR #551, opened 2025-02-19
Improve register performance

colby-nyce:


klingaard (approved):

furuame:
Wondering why not use `sparta_assert`?

klingaard:
> Wondering why not use `sparta_assert`?

`assert` in release mode of simulation builds are removed allowing the compiler to inline the function (removes the implicit branch).

PR #553, opened 2025-02-20
Make FastList compatible with C++23

bdutro:
Applies the same fixes from https://github.com/sparcians/map/pull/498 to FastList.


klingaard (approved):
Ugh, can't they leave well-enough alone?!?

PR #555, opened 2025-03-12
New memory type: CachedMemory

klingaard:
CachedMemory is a new type of memory object that allows a user to write memory to a temporary
location in a speculative manner.  When that  memory is guaranteed to be committed (written to 
main memory or downstream memory), the modeler can request CachedMemory to force it out.

If the memory is _not_ to be pushed out (flushing for example), then the modeler can "drop" that memory
update and CachedMemory will restore the previous value.

Use case: data-driven performance modeling where there are speculative stores (storing to a local store commit buffer)
and there dependent loads on that data.  If the store is no longer speculative, the store is officially committed.  If it's 
flushed, old data can be restored and downstream memory (main memory/IO devices, etc) will never see the data.


sparta/CMakeLists.txt
```diff
@@ -202,6 +198,10 @@ elseif (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
   # using Visual Studio C++
 endif ()
 
+Set(CMAKE_CXX_FLAGS_DEBUG          "-O0 -g -U_FORTIFY_SOURCE")
```
colby-nyce, line 201: Nit: set or SET for cmake, not Set

sparta/sparta/memory/CachedMemory.hpp
```diff
@@ -118,15 +118,26 @@ namespace sparta::memory
...
      * When the write is committed by a functional model (like in a
      * cosimulation environment), the data is pushed downstream.
      *
+     * If the memory is _not_ to be pushed out (flushing for example),
+     * the modeler can "drop" that memory update and CachedMemory will
+     * restore the previous value.
+     *
+     * Use case: data-driven performance modeling where there are
+     * speculative stores (storing to a local store commit buffer) and
+     * there dependent loads on that data.  If the store is no longer
```
colby-nyce, line 136: there "are" dependent loads?

sparta/sparta/memory/CachedMemory.hpp
```diff
@@ -0,0 +1,487 @@
...
+        StoreData(uint64_t        write_id,
+                  addr_t          paddr,
+                  size_t          size,
+                  const uint8_t * data,
+                  const void    * in_supplement,
+                  void          * out_supplement) :
+            write_id_(write_id),
+            paddr_(paddr),
+            size_(size),
+            storeed_value_(size),
```
colby-nyce, line 70: stored_value_ ?

sparta/sparta/memory/CachedMemory.hpp
```diff
@@ -0,0 +1,487 @@
...
+        const addr_t         paddr_;  //!< Physical address
+        const size_t         size_;   //!< doubleword=8, word=4, halfword=2, byte=1
+        std::vector<uint8_t> storeed_value_;
+        std::vector<uint8_t> prev_value_;
+    };
+
+    inline std::ostream & operator<<(std::ostream & os, const StoreData & mstore)
+    {
+        os << std::hex << "wid:" << mstore.getWriteID() << " pa:" << mstore.getPAddr()
+           << " size:" << mstore.getSize() << std::dec;
```
colby-nyce, line 112: Is this printing the size in hex?

colby-nyce (approved):

sparta/CMakeLists.txt
klingaard: easy to fix

sparta/sparta/memory/CachedMemory.hpp
klingaard: it's a new word.  :smile: 

Fixed

sparta/sparta/memory/CachedMemory.hpp
klingaard: good catch.  fixed

sparta/test/CachedMemory/CachedMemory_test.cpp
```diff
@@ -0,0 +1,560 @@
...
+    ////////////////////////////////////////////////////////////////////////////////
+    // Test dropping of a outstanding write and that is restores
+    //
+    // First, test dropping of a write not performed on this
+    // CachedMemory.  Second, drop write from the previous tests (at
+    // 0x2000).  Memory should be restored to previous value.
+    const sparta::memory::StoreData bad_write(0, paddr_0x1000, 8,
+                                              ffffffff_data.data_ptr.data(),
+                                              nullptr, nullptr);
+    EXPECT_THROW  (test_system.cached_mem_core0.dropWrite(bad_write));
```
kathlenemagnus, line 223: Remove extra spaces

kathlenemagnus (approved):
Great tests. Looks good to me!

sparta/test/CachedMemory/CachedMemory_test.cpp
klingaard: done

PR #565, opened 2025-05-07
Get all SimDB-exported CSV reports to match MAP v2 formatters exactly

colby-nyce:
tl;dr - In total, 61 different tests all match exactly for the csv and csv_cumulative formats.

I used the "create_tests.py" script to create a much larger tests.yaml -- the script turns every sparta_named_test into a yaml definition that the test suite can use (basically scripts a new generation of tests.yaml instead of the hand rolled file I wrote with 18 or so tests in it).

### DIFF
--- a/sparta/src/ReportDescriptor.cpp
+++ b/sparta/src/ReportDescriptor.cpp
@@ -318,6 +318,19 @@ void ReportDescriptor::sweepSimDbStats_()
 #endif
 }
 
+void ReportDescriptor::skipSimDbStats_()
+{
+#if SIMDB_ENABLED
+    if (db_mgr_ == nullptr || skipped_annotator_ == nullptr) {
+        return;
+    }
+
+    const std::string annotation = dest_file + "_skipped_anno_" + skipped_annotator_->currentAnnotation();
+    const auto tick = scheduler_->getCurrentTick();
+    db_mgr_->getCollectionMgr()->sweep("root", tick, annotation, true);
+#endif
+}
+
 report::format::BaseFormatter* ReportDescriptor::addInstantiation(Report* r,
                                                                   Simulation* sim,
                                                                   std::ostream* out)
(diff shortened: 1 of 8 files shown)


sparta/src/ReportDescriptor.cpp
```diff
@@ -186,16 +186,16 @@ std::shared_ptr<statistics::StreamNode> ReportDescriptor::createRootStatisticsSt
     return nullptr;
 }
 
-void ReportDescriptor::configSimDbReports(simdb::DatabaseManager* db_mgr, RootTreeNode* root)
+int ReportDescriptor::configSimDbReports(simdb::DatabaseManager* db_mgr, RootTreeNode* root)
```
klingaard, line 189: What does the return value mean?

klingaard (approved):

sparta/src/ReportDescriptor.cpp
colby-nyce:
```
            /*!
             * \brief Write all metadata about our report (and its subreports
             * and statistics) to SimDB.
             * 
             * \return Returns the database ID of the report descriptor in the 
             * ReportDescriptors table. Returns 0 if:
             *    - SIMDB_ENABLED is false
             *    - this descriptor is not enabled
             *    - this descriptor has no reports
             */
            int configSimDbReports(simdb::DatabaseManager* db_mgr, RootTreeNode* root);
```

sparta/src/ReportRepository.cpp
```diff
@@ -110,6 +110,37 @@ class Directory
         reports_.emplace_back(report.release());
     }
 
+    /*!
+     * \brief Write all metadata about our report (and its subreports
+     * and statistics) to SimDB.
+     */
+    void configSimDbReports(simdb::DatabaseManager* db_mgr, RootTreeNode* root)
+    {
+    #if SIMDB_ENABLED
```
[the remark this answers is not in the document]
colby-nyce: Good catch, fixed

PR #566, opened 2025-05-08
Rewrite the SimDB report verif tester

colby-nyce:
This single test script will only parse CoreModel/CMakeLists.txt and do everything from there. Previous implementation was too clunky. The output of this new tester currently looks like this:

```
Final results:

Format   Passed   Failed
-----------------------------
csv      86       0       
json     0        24      
html     0        6       
txt      0        19
```


sparta/scripts/simdb/run_report_verif_suite.py
```diff
@@ -0,0 +1,490 @@
...
+
+            if not os.path.exists(simdb_report):
+                self.__WriteToTestLog(f"SimDB report {report} not found.")
+                failing_reports.append(report)
+                continue
+
+            # Compare the two reports.
+            comparator = GetComparator(report)
+            if not comparator.Compare(baseline_report, simdb_report):
+                self.__WriteToTestLog(f"Baseline report {report} does not match SimDB report. Test will fail.")
```
kathlenemagnus, line 259: Does it make sense to report a different type of failure if `GetComparator` returns an `UnsupportedComparator`?

kathlenemagnus (approved):
Neat!

sparta/scripts/simdb/run_report_verif_suite.py
colby-nyce: Sure! I can add a column to the results like this:

```
Final results:

Format   Passed   Failed   NoCompare
csv      86       0        0
json     0        24       0
html     0        6        0
txt      0        19       0
gnuplot  0        0        3
```

colby-nyce: The results now look like this -- html is the only one in any CMakeLists.txt test which we have to discuss how to compare (if we even support this format at all in MAP v3):

```
Final results:

Format   Passed   Failed   NoCompare
txt      0        19       0       
html     0        0        6       
csv      86       0        0       
json     0        24       0
```

PR #568, opened 2025-05-12
SimDB-exported json reports

colby-nyce:
These changes bring up all the JSON formats to match legacy exactly (except the js_json format, coming soon).

Here is where we stand now:

```
Format       Passed   Failed   NoCompare
csv          86       0        0       
html         0        0        6       
js_json      0        2        0       
json         7        0        0       
json_detail  5        0        0       
json_reduced 8        0        0       
txt          0        19       0
```

### DIFF
--- a/sparta/sparta/report/format/JSON_detail.hpp
+++ b/sparta/sparta/report/format/JSON_detail.hpp
@@ -16,7 +16,6 @@
 
 #include "sparta/report/format/BaseOstreamFormatter.hpp"
 #include "sparta/report/format/DetailInfoData.hpp"
-#include "sparta/report/db/DatabaseContextCounter.hpp"
 #include "sparta/utils/SpartaException.hpp"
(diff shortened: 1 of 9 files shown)


sparta/sparta/report/format/JSON_detail.hpp
```diff
@@ -211,6 +213,52 @@ class JSON_detail : public BaseOstreamFormatter
             local_name = p_name + "." + flattenReportName(r->getName());
         }
 
+        auto extract_stat = [&local_name](const statistics::stat_pair_t & si) {
```
colby-nyce, line 216: Note that this code was accidentally removed from Sparta when SimDB was removed. I verified that all json_detail reports are identical when comparing before-and-after SimDB removal (to make sure that I put this code back correctly) and everything passes in the SimDB-vs-legacy json_detail tests too.

klingaard (approved):

PR #571, opened 2025-05-14
Get the SimDB stats_mapping report to match legacy

colby-nyce:
This only affects the stats_mapping test which passes:

```
Format        Passed   Failed   NoCompare
csv           86       0        0       
html          0        0        6       
js_json       0        2        0       
json          9        0        0       
json_detail   5        0        0       
json_reduced  8        0        0       
stats_mapping 1        0        0       
txt           23       0        0
```

### DIFF
--- a/sparta/src/StatsMapping.cpp
+++ b/sparta/src/StatsMapping.cpp
@@ -20,7 +20,7 @@ namespace format {
 
 void createStatsMappingForReport(const Report * r,
                                  const std::string & prefix,
-                                 std::unordered_map<std::string, std::string> & mapping)
+                                 std::map<std::string, std::string> & mapping)
 {
     for (const auto & si : r->getStatistics()) {
         if (!si.first.empty()) {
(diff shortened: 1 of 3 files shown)


sparta/src/StatsMapping.cpp
```diff
@@ -54,6 +49,7 @@ void StatsMapping::writeHeaderToStream_(std::ostream & out) const
     rapidjson::Value stats2headers;
     stats2headers.SetObject();
 
+    std::map<std::string, std::string> reverse_mapping;
```
kathlenemagnus, line 52: Why is it necessary to invert the map?

kathlenemagnus (approved):

sparta/src/StatsMapping.cpp
colbynyce-mips: This is just the format that was requested back in the original implementation for faster lookup in both directions (abbreviated below):

```
{
    "Column-header-to-statistic": {
        "scheduler.seconds": "scheduler.stats.seconds",
        "scheduler.ticks": "scheduler.stats.ticks"
    },
    "Statistic-to-column-header": {
        "scheduler.stats.seconds": "scheduler.seconds",
        "scheduler.stats.ticks": "scheduler.ticks"
    },
    "report_metadata": {
        "report_format": "stats_mapping"
    }
}
```

PR #572, opened 2025-05-14
Refactor the SimDB exporters

colby-nyce:


colby-nyce:
I found myself duplicating code while I was writing the js_json exporter, so here's the refactor PR while I finish js_json.

kathlenemagnus (approved):

PR #579, opened 2025-05-28
Use parameter validation to check TreeNode extension values

colby-nyce:
These changes replace the hand-rolled validation code in validateParameter() and use the "official" Sparta parameter validation instead.


klingaard (approved):

kathlenemagnus-mips (commented):
Thanks @colby-nyce!

PR #581, opened 2025-05-29
SimDB reports perf improvements

colby-nyce:
The main goal of these changes is to get v2.1 SimDB reports to be faster and smaller than legacy reports. The biggest issue here was the `sparta.db` size when using timeseries reports, previously `33% larger than formatted reports`.

The performance benchmark used five different timeseries reports with 1M instructions (sparta_core_example).

```
                Time    Size
Legacy          6.7s    36MB
SimDB before    5.8s    48MB   (33% larger)
SimDB after     5.5s    28MB   (22% smaller)
```

So the performance we are left with (compared to legacy) is `18% faster, 22% smaller`

All tests are passing:

```
Format        Passed   Failed   NoCompare
csv           86       0        0       
html          0        6        0       
js_json       3        0        0       
json          9        0        0       
json_detail   5        0        0       
json_reduced  8        0        0       
stats_mapping 1        0        0       
txt           23       0        0
```

Report export performance is not explicitly addressed in this PR, though the `simdb_export` script got slower by 2% due to the schema changes to make the `sparta.db` much smaller. Export perf will be improved in a later PR.

I'd recommend reviewing the C++/schema changes first, then the python code.

Here are the supporting SimDB changes (already in main):
https://github.com/colby-nyce/simdb/commit/5d6db040c0c0d1d2a60ed01ffb84d625aa5fa9db

### DIFF
--- a/sparta/scripts/simdb/exporters/export_csv_report.py
+++ b/sparta/scripts/simdb/exporters/export_csv_report.py
@@ -89,17 +89,7 @@ class CSVReportExporter:
                 if is_compressed:
                     data = zlib.decompress(data)
 
-                # The data values are stored as:
-                #   [elem_id(u16), value(double), elem_id(u16), value(double), ...]
-                #
-                # We only care about the values, not the element IDs. Everything in these records
-                # is already in the order we want, meaning that the values line up with the stat
-                # headers we already wrote out.
-                #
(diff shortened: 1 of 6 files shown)


klingaard (approved):
Nice!

PR #585, opened 2025-06-26
Fix compiler errors with Clang 20 and GCC 15

bdutro:


bdutro:
> Will need to merge this in with `map_v2` when you do so.

How would you prefer I do that? Just merge master -> map_v2 or a cherry-pick?

bdutro:
Looks like map_v2 isn't fully up to date with master, so I'll just cherry-pick this fix in.

PR #587, opened 2025-06-27
[map_v2] More Sparta build improvements

bdutro:
Fix failure when trying to build against an HDF5 library with MPI support enabled
Maximize RapidJSON compatibility by using both possible versions of the include path CMake variable (`RapidJSON_INCLUDE_DIRS` and `RAPIDJSON_INCLUDE_DIRS`)
Fix compiler error with Clang 20+ and GCC 15+

### DIFF
--- a/sparta/simdb/cmake/simdb-config.cmake
+++ b/sparta/simdb/cmake/simdb-config.cmake
@@ -11,3 +11,9 @@ enable_language(C)
 find_package(HDF5 1.10 REQUIRED)
 
 set(SimDB_LIBS simdb ${HDF5_LIBRARIES} sqlite3 z pthread)
+
+# If HDF5 is built with MPI support, we also need to add the MPI include dirs and link against the MPI library
+if(HDF5_IS_PARALLEL)
+    find_package(MPI REQUIRED)
+    list(APPEND SimDB_LIBS MPI::MPI_CXX)
+endif()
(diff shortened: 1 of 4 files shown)


colby-nyce (approved):

sparta/simdb/cmake/simdb-config.cmake
```diff
@@ -11,3 +11,9 @@ enable_language(C)
 find_package(HDF5 1.10 REQUIRED)
 
 set(SimDB_LIBS simdb ${HDF5_LIBRARIES} sqlite3 z pthread)
+
+# If HDF5 is built with MPI support, we also need to add the MPI include dirs and link against the MPI library
+if(HDF5_IS_PARALLEL)
+    find_package(MPI REQUIRED)
+    list(APPEND SimDB_LIBS MPI::MPI_CXX)
```
colby-nyce, line 18: What was the reason to add it to SimDB_LIBS? This would have to go through the SimDB repo:
https://github.com/colby-nyce/simdb

But I'd like to know why it was needed.

bdutro: I checked the new SimDB repo and it looks like it doesn't directly interface with HDF5 anymore.

colby-nyce: What's the verdict on requiring this in SimDB? If I go back and remove it from map_v2, will Sparta fail to build?

bdutro: As long as nothing in SimDB directly includes or links against HDF5 then this isn't needed.

PR #596, opened 2025-07-25
map_v2.1.0 (approach #1)

colby-nyce:
This is a good cutoff point for map_v2.1+ reports, plus everything we need for Pegasus CoSim and the profiler. It would be nice to get these changes in as well:

https://github.com/sparcians/map/pull/595

Although cosim has workarounds without it.


colby-nyce:
@klingaard I believe we can back out the last commit and do something else to get the CI workflow to work with ssh if https is not okay for the simdb submodule.

klingaard:
Hmmm.... what do you think about making a branch off of master called map_v2.1?  Merging this with map_v2 will break all those folks that are using (and updating from) map_v2.

colby-nyce:
I figured this was okay for map_v2 since the use of SimDB is opt-in only (`--enable-simdb-reports`)

If anybody was expecting the old SimDB headers to be there, then yes this would break their build and should be map_v2.1

Is that last point the only concern? If not - what else could break?

colby-nyce:
I'm going to open a separate PR with Knute's suggestion.

colby-nyce:
Closing in favor of map_v2.1 branch

PR #597, opened 2025-07-25
map_v2.1.0 (approach #2)

colby-nyce:
This PR is different from [#596](https://github.com/sparcians/map/pull/596) in that it is cherry picking SimDB commits to a new branch "map_v2.1" so as not to break anybody's workflow or builds.


colby-nyce:
Now that all checks are passing, I'll take the required changes to master first, then return to this PR to release map_v2.1

PR #603, opened 2025-08-08
Fix memory leak from the payload event proxy

chengdelinTT:
Memory leaks can occur when an event reclaims a proxy containing a payload that is a structure with an internal pointer, rather than a pointer itself. The event's reset mechanism only handles pointer-type payloads, which means structures with pointers are not properly cleared and remain embedded within the event. Consequently, even after the scheduled event completes and the object is no longer referenced elsewhere, it persists until the simulation's conclusion.

To mitigate this issue, we propose transforming the payload type from `DataT` to `std::optional<DataT>`. By consistently resetting the payload to `std::nullopt` during proxy reclamation, we ensure that freed proxies no longer retain pointer references. This approach allows the allocator to release any objects previously contained in the PhasedPayloadEvent's payload once they are no longer referenced in the model.

### DIFF
--- a/sparta/sparta/events/PhasedPayloadEvent.hpp
+++ b/sparta/sparta/events/PhasedPayloadEvent.hpp
@@ -221,7 +221,11 @@ namespace sparta
         }
 
         //! Destroy!
-        virtual ~PhasedPayloadEvent() {}
+        virtual ~PhasedPayloadEvent() {
+            for(PayloadDeliveringProxy * proxy : inflight_pl_) {
+                std::destroy_at(proxy->payload_);
+            }
+        }
 
         //! No assignments, no copies
         PhasedPayloadEvent<DataT> & operator=(const PhasedPayloadEvent<DataT> &) = delete;
--- a/sparta/test/Events/PayloadEventMemLeakTest.cpp
+++ b/sparta/test/Events/PayloadEventMemLeakTest.cpp
@@ -46,6 +46,9 @@ int main()
     // The payload event that send a data containing a shared pointer
     sparta::PayloadEvent<PointerWrapper> pld_ptr_wrapper_event(&event_set, "pld_ptr_wrapper_event", CREATE_SPARTA_HANDLER_WITH_DATA_WITH_OBJ(EventHandler, &ev_handler, handler, PointerWrapper), 0);
 
+    // unique pointer to the event
+    std::unique_ptr<sparta::PayloadEvent<sparta::SpartaSharedPointer<uint32_t>>> pld_event_ptr = std::make_unique<sparta::PayloadEvent<sparta::SpartaSharedPointer<uint32_t>>>(&event_set, "pld_event_ptr", CREATE_SPARTA_HANDLER_WITH_DATA_WITH_OBJ(EventHandler, &ev_handler, handler, sparta::SpartaSharedPointer<uint32_t>), 0);
+
     // Monitor the memory usage of the shared pointer
     sparta::SpartaSharedPointerAllocator<uint32_t> int_shared_pointer_allocator(1, 1);
 
@@ -73,7 +76,15 @@ int main()
     // The shared pointer is not stored anywhere other than the event payload
     EXPECT_FALSE(int_shared_pointer_allocator.hasOutstandingObjects());  
 
+    pld_event_ptr->preparePayload(sparta::allocate_sparta_shared_pointer<uint32_t>(int_shared_pointer_allocator, 0))->schedule(3);
+
     rtn.enterTeardown();
+
+    // check the payload of the outstanding event is destroyed
+    pld_event_ptr.reset();
+    EXPECT_FALSE(int_shared_pointer_allocator.hasOutstandingObjects());  
+
+    
 
     REPORT_ERROR;
     return ERROR_CODE;


klingaard:
Thank you for your contribution!  I see the issue you're trying to solve.  `std::optional` is a great solution, but I worry about the addition performance implications it adds for the billions of PLEs that are created, accessed, and then destroyed in some of our large simulations.

Specifically, for each PLE destruction, there's this `if` statement in `std::optional`:
```c++
constexpr void                                          
_M_reset() noexcept                                     
{                                                       
  if (this->_M_engaged)                                 
    _M_destroy();                                       
  else 
    this->_M_engaged = false;
}
```

And for every retrieval of the optional value, there's an implicit `if` statement.  Could cost quite a bit of time...

Curious, did you consider just adding an explicit destruction call?  Something like this (not tested):
```c++
            // Destroy payload
            void PayloadDeliveringProxy::destroyPayload_() {
                sparta_assert(scheduled_ == false);
                if constexpr(MetaStruct::is_any_pointer<DataT>::value) {
                    payload_ = nullptr;
                    return;
                }
                if constexpr(std::is_class_v<DataT>) {
                    payload_.~DataT();
                }
            }

```
and just call that?
```c++
        void reclaimProxy_(typename ProxyInflightList::iterator & pl_location) {
            if (SPARTA_EXPECT_FALSE(pl_location == inflight_pl_.end())) {
                return;
            }
            (*pl_location)->destroyPayload_();
            free_pl_[free_idx_++] = *pl_location;
            inflight_pl_.erase(pl_location);
            pl_location = inflight_pl_.end();
        }
```

Let me know what you think.

klingaard:
BTW, THANK YOU for adding a test!

sparta/sparta/events/GlobalEvent.hpp
```diff
@@ -58,7 +58,9 @@ namespace sparta
 
         const SchedulingPhase & getSchedulingPhase() const { return phase_; }
 
-        ~GlobalEventProxy() = default;
+        ~GlobalEventProxy(){
+            ev_handler_.reset();
```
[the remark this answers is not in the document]
chengdelinTT: I just realize that I did not run the full CI test. I only ran `make regress` instead of all 3 regress. My bad and I'll figure it out.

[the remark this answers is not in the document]
klingaard: Looks like a double-delete going on for concrete objects.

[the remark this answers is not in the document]
klingaard: One possible fix: delete the destructor for PLEProxy

chengdelinTT:
I feel like the problem is because we call the destructor twice on the same object. One when the proxy releases (by this PR change), and the other one is when the old payload is replaced by the new one.

I think we can use `std::unique_ptr<DataT>` to store the payload.  We still create a new payload object but use an unique pointer to reference. When releasing the proxy, PLE can set the proxy to `nullptr` to release the proxy, just like `std::nullopt` for `std::optional<DataT>`.  Therefore we do not manually called the destructor (2nd approach) and also does not have the additional check (1st approach through `std::optional<DataT>`).

Please let me know if you feel OK about this approach. Thanks!

sparta/sparta/events/PhasedPayloadEvent.hpp
```diff
@@ -127,13 +139,14 @@ namespace sparta
...
-                target_consumer_event_handler_((const void*)&payload_);
+                target_consumer_event_handler_((const void*)payload_);
                 reclaim_();
             }
 
             PhasedPayloadEvent<DataT> * parent_ = nullptr;
             const SpartaHandler         target_consumer_event_handler_;
-            DataT                       payload_;
+            DataT *                     payload_;
+            std::byte                   payload_storage_[sizeof(DataT)];
```
klingaard, line 149: You might need to align this memory just to be safe.  See the example: https://en.cppreference.com/w/cpp/memory/construct_at.html

klingaard (changes requested):
Hate to do this to ya, but I think there's still a bug in the implementation.  What if there are outstanding payloads and simulation ends?  In the past, the destructor the proxy would be called and would delete the payload.  But now, it's not deleted at all. 

I think in the destructor for the PLE, you need to walk outstanding proxies and call the destroy method

sparta/sparta/events/PhasedPayloadEvent.hpp
```diff
@@ -104,12 +104,24 @@ namespace sparta
             // Set a payload for a delayed delivery
             void setPayload_(const DataT & pl) {
                 sparta_assert(scheduled_ == false);
-                payload_ = pl;
+                payload_ = new (&payload_storage_) DataT(pl);
```
klingaard, line 107: Also, to be safer, consider using `std::construct_at`

sparta/sparta/events/PhasedPayloadEvent.hpp
```diff
@@ -104,12 +104,24 @@ namespace sparta
...
+
+            // Destroy payload
+            void destroyPayload_() {
+                sparta_assert(scheduled_ == false);
+                if constexpr(MetaStruct::is_any_pointer<DataT>::value) {
+                    *payload_ = nullptr;
+                    return;
+                }
+                if constexpr(std::is_class_v<DataT>) {
+                    payload_->~DataT();
```
klingaard, line 118: Consider using `std::destroy_at`

sparta/sparta/events/PhasedPayloadEvent.hpp
chengdelinTT: Does sparta still target in C++17? `std::construct_at` is C++20 feature according to https://en.cppreference.com/w/cpp/memory/construct_at.html .

chengdelinTT:
> What if there are outstanding payloads and simulation ends? 

Good catch! Added the destructor that destroy the outstanding proxy and also the corresponding test. I use `std::destory_at` instead of `proxy->destroyPayload_()` since we probably don't want to check the assertion inside the function when destructing the PLE.

sparta/sparta/events/PhasedPayloadEvent.hpp
chengdelinTT: Done TKX

klingaard:
I'll merge this tomorrow if all is well.

chengdelinTT:
Gotcha. Thanks for your effort and sorry for the tight deadline.

PR #605, opened 2025-08-26
Database-backed checkpointer

colby-nyce:
This checkpointer is not done, but I'm having some writer's block / difficulty keeping all this in my head since it has quite a few moving pieces to get right. So I'd like to just open the PR at this point and try to share the cognitive load a little :)


sparta/sparta/serialization/checkpoint/Checkpoint.hpp
```diff
@@ -255,23 +180,12 @@ namespace sparta::serialization::checkpoint
         }
 
     private:
-
-        const tick_t tick_; //!< Tick number for this checkpoint.
+        tick_t tick_;         //!< Tick number for this checkpoint.
```
kathlenemagnus, line 183: Why would the tick be changed?

sparta/sparta/serialization/checkpoint/FastCheckpointer.hpp
```diff
@@ -379,13 +393,29 @@ namespace sparta::serialization::checkpoint
             return d;
         }
 
+        /*!
+         * \brief Finds a checkpoint by its ID
+         * \param id ID of checkpoint to find. Guaranteed not to be flagged as
+         * deleted
+         * \return Checkpoint with ID of \a id if found or nullptr if not found
+         */
+        checkpoint_type* findCheckpoint(chkpt_id_t id) noexcept {
```
kathlenemagnus, line 402: Should this return a const pointer so the checkpoint can't be modified?

sparta/sparta/serialization/checkpoint/DatabaseCheckpoint.hpp
```diff
@@ -0,0 +1,304 @@
...
+
+        /*!
+         * \brief Can this checkpoint be deleted
+         * Cannot be deleted if:
+         * \li This checkpoint has any ancestors which are not deletable and not snapshots
+         * \li This checkpoint was not flagged for deletion with flagDeleted
+         * \warning This is a recursive search of a checkpoint tree which has potentially many
+         * branches and could have high time cost
+         */
+        bool canDelete() const noexcept;
```
kathlenemagnus, line 165: In what scenarios can a checkpoint not be deleted? Is this mainly for `DeltaCheckpoint` support?

colby-nyce:
Note that this is a draft and should not be reviewed until switched back.

sparta/sparta/serialization/checkpoint/Checkpoint.hpp
```diff
@@ -160,93 +148,30 @@ namespace sparta::serialization::checkpoint
...
-            }
-            if(next->getPrev() != this){
-                throw CheckpointError("Attempting to remove a next checkpoint whose previous pointer is not 'this'");
-            }
-            auto itr = std::find(nexts_.begin(), nexts_.end(), next);
-            if(itr == nexts_.end()){
-                throw CheckpointError("Next argument was not present in this checkpoint's nexts_ list. Cannot remove");
-            }
-            nexts_.erase(itr);
+        virtual std::string getDeletedRepr() const {
```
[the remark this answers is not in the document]
colby-nyce: It's called by the checkpointer classes. I rather not make it private because then I'd have to make the checkpointers friends of the Checkpoint base class. Right now, FastCheckpointer is a friend of DeltaCheckpoint, and DatabaseCheckpointer is a friend of DatabaseCheckpoint. They are used together 1-to-1 so I'm okay with that friendship, but making all checkpointers friends of CheckpointBase doesn't seem right to me.

sparta/sparta/serialization/checkpoint/DatabaseCheckpoint.hpp
colby-nyce: Yes it was really for `DeltaCheckpoint`. Explicit deletion of database checkpoints has been removed for performance reasons. So `canDelete()` and `getDeletedID()` have been removed and left in `DeltaCheckpoint` only.

sparta/sparta/serialization/checkpoint/Checkpoint.hpp
colby-nyce: The `tick_` has to be non-const to support the `DatabaseCheckpoint` subclass, which needs to be recreated from raw bytes using `boost::serialization`. It was const before since the `DeltaCheckpoint` is only ever held in memory so the tick was just given to the ctor.

sparta/sparta/serialization/checkpoint/Checkpointer.hpp
```diff
@@ -831,7 +753,7 @@ namespace sparta::serialization::checkpoint
         /*!
          * \brief Most recent checkpoint created or loaded
          */
-        Checkpoint* current_;
+        CheckpointBase* current_;
```
knute-mips, line 756: Should these be initialize to nullptr?

PR #612, opened 2025-10-16
DatabaseCheckpointer performance optimizations

colby-nyce:
I added a variety of utilities/APIs to SimDB to get `loadCheckpoint()` numbers way down. These changes result in most checkpoints being loadable from the cache directly (blazing fast) and now from the pipeline task's queues (very fast). Similar optimizations are done for checkpoint deletion.

When not found in the cache or pipeline, the fallback plan of just querying the database can still take 1-8ms or so. Without these changes in this PR, most use cases fell into that bucket. But retrieving directly from the pipeline when we are able to (even with "undoing" `boost::serialization` and zlib) takes around **20 microseconds** now to perform `loadCheckpoint()`.

The CoSim event pipeline (500 microseconds) used to set the standard, but is now the bottleneck =D

### DIFF
--- a/sparta/sparta/app/simdb/ReportStatsCollector.hpp
+++ b/sparta/sparta/app/simdb/ReportStatsCollector.hpp
@@ -39,8 +39,7 @@ public:
 
     static void defineSchema(simdb::Schema&);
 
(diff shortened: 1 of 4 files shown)


colby-nyce:
Moving this to a draft PR until I figure out the MacOS failure.

PR #623, opened 2025-11-12
Added a class Checkpointable that integrates with Sparta's CP mechanism

klingaard:
Class that allows a modeler to integrate local variables in simulation into Sparta's checkpoint mechanism.


colby-nyce (approved):
Nice! Very simple and clean design.

PR #651, opened 2026-02-17
Fix all known issues with tree node extensions

colby-nyce:
These changes passed all regressions in Sparta and all of our internal simulators (with a few unavoidable changes). Olympia will also have to change a bit (a couple `getExtension()` calls need to be `createExtension()`).

I also grepped all the log files from the regression tests looking for "unbound" to make sure that it doesn't even occur as a warning.

### DIFF
--- a/sparta/src/CommandLineSimulator.cpp
+++ b/sparta/src/CommandLineSimulator.cpp
@@ -1696,6 +1696,10 @@ bool CommandLineSimulator::parse(int argc,
     }
 
     if (!global_simdb_file_.empty()) {
+        if (global_simdb_file_ == "autogen") {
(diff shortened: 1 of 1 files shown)


sparta/sparta/app/SimulationConfiguration.hpp
```diff
@@ -479,14 +479,29 @@ class SimulationConfiguration
...
                 global_simdb_file_ = db_file;
             }
+
+            if (reuse)
+            {
+                reuseDatabase(db_file);
+            }
+        }
+
+        void reuseDatabase(const std::string & db_file)
```
colby-nyce: Nothing to do with extensions. Needed for post-simulation replayer (connect to existing database).

PR #677, opened 2026-06-22
Add std namespace scope to isinf and isnan functions

elliottgurrola-mips:
Add std:: namespace to isinf and isnan fucntion calls as I was getting a compile error while using c++ (conda-forge gcc 15.2.0-7) 15.2.0


klingaard (approved):
Geez how big was the dust cloud when you blew on that code?  🤣

elliottgurrola-mips:
I need to be fast merging this before Colby sees my PR.

PR #679, opened 2026-07-21
Argos collector integration for map_v3

colby-nyce:


sparta/sparta/collection/BitBucket.hpp
```diff
@@ -0,0 +1,229 @@
...
+//! \class BitBucket
+//! \brief Base class for all collectable bit buckets (scalars, sparse/contig iterables).
+//! We use this class so that we can gather data from e.g. Pair objects without requiring
+//! that they hold onto their own collected bytes/values. When data values become available,
+//! they go directly into the BitBucket and are organized into data structures that SimDB
+//! expects (depends on the collected type).
+class BitBucket
+{
+public:
+    explicit BitBucket(simdb::TinyStrings<>* tiny_strings, simdb::argos::EnumInspector* enum_inspector) :
```
klingaard: `explicit` not needed

sparta/sparta/collection/BitBucket.hpp
```diff
@@ -0,0 +1,229 @@
...
+        // or bools as uint8_t)
+        else if (std::is_trivial_v<T> && std::is_standard_layout_v<T>) {
+            writeField(&val, sizeof(T), field_id);
+            return true;
+        }
+
+        // Cannot collect this field. Let the caller decide what to do.
+        else {
+            return false;
+        }
```
klingaard, line 86: Can you not static_assert?

sparta/sparta/collection/BitBucket.hpp
```diff
@@ -0,0 +1,229 @@
...
+            return writeField(tiny_strings_->getStringID(val), field_id);
+        }
+
+        // Write enums as their underlying integer type.
+        else if constexpr (std::is_enum_v<T>) {
+            enum_inspector_->inspect(val);
+            using underlying_t = std::underlying_type_t<T>;
+            const underlying_t enum_int = static_cast<underlying_t>(val);
+            return writeField(enum_int, field_id);
+        }
```
klingaard, line 48: Do you have a mechanism (haven't looked at the entire PR) to describe the enum for Argos?

sparta/sparta/collection/BitBucket.hpp
```diff
@@ -0,0 +1,229 @@
...
+
+        // Write struct/class fields that provide exactly one cast-to-POD operator
+        else if constexpr (simdb::type_traits::is_pod_convertible_v<T> && !std::is_enum_v<T> &&
+                          (!std::is_trivial_v<T> || !std::is_standard_layout_v<T>)) {
+            using converted_t = simdb::type_traits::pod_convertible_t<T>();
+            static_assert(std::is_trivial_v<converted_t> && std::is_standard_layout_v<converted_t>);
+            auto converted_val = static_cast<converted_t>(val);
+            return writeField(converted_val, field_id);
```
klingaard: Does this mean that my non-POD types must be decomposable to a simple, singular POD representation?

sparta/sparta/collection/BitBucket.hpp
```diff
@@ -0,0 +1,229 @@
...
+        return enum_inspector_;
+    }
+
+private:
+    simdb::TinyStrings<>* tiny_strings_ = nullptr;
+    simdb::argos::EnumInspector* enum_inspector_ = nullptr;
+};
+
+//! BitBucket implementation for Collectable objects (whether "standalone"
+//! or inside an IterableCollector).
```
klingaard: Is this the description for BitBucket or CollectableBitBucket?

sparta/sparta/collection/BitBucket.hpp
klingaard, line 1: I'm a bit confused on the design.  There are two methods called `writeField`.  The templatized version seems to be recursive devolving the type?

Should the virtual `writeField` be protected?

sparta/sparta/collection/BitBucket.hpp
```diff
@@ -0,0 +1,229 @@
...
+public:
+    using BitBucket::BitBucket;
+
+    void clear() override final {
+        buffer_.clear();
+        buffer_.reserve(bytes_per_pass_);
+        bytes_per_pass_ = 0;
+    }
+
+    void writeField(const void* data, uint32_t bytes, uint32_t) override final {
```
klingaard: Should modelers be allowed to call this?  Should they always call the template version?

sparta/sparta/collection/BitBucket.hpp
```diff
@@ -0,0 +1,229 @@
...
+
+    void clear() override final {
+        active_bin_idx_.clearValid();
+        all_bin_idxs_.clear();
+        all_bin_idxs_.reserve(capacity_);
+    }
+
+    void setActiveBinIdx(uint32_t bin_idx) {
+        assert(bin_idx <= UINT16_MAX);
+        assert(bin_idx < capacity_);
```
klingaard: There's a good chance this will always be compiled non-debug in the framework.  `sparta_assert`?

sparta/sparta/collection/BitBucket.hpp
```diff
@@ -0,0 +1,229 @@
...
+        active_bin_idx_ = static_cast<uint16_t>(bin_idx);
+        all_bin_idxs_.emplace_back(bin_idx);
+    }
+
+    void writeField(const void* data, uint32_t bytes, uint32_t field_id) override final {
+        auto& bin_bucket = bin_buckets_[active_bin_idx_.getValue()];
```
klingaard: `bit_buckets_.at`?

sparta/sparta/collection/BitBucket.hpp
```diff
@@ -0,0 +1,229 @@
...
+public:
+    IterableCollectorBitBucket(simdb::TinyStrings<>* tiny_strings, simdb::argos::EnumInspector* enum_inspector, size_t capacity)
+        : BitBucket(tiny_strings, enum_inspector)
+        , capacity_(capacity)
+    {
+        while (capacity--)
+        {
+            bin_buckets_.emplace_back(std::make_unique<CollectableBitBucket>(tiny_strings, enum_inspector));
+        }
+    }
```
klingaard, line 166: You don't call clear() here.

sparta/sparta/collection/BitBucket.hpp
```diff
@@ -0,0 +1,229 @@
...
+            bin_bucket->writeTo(bin_buffer);
+        }
+
+        entry_point->setContigContainerBinBytes(std::move(all_bin_bytes_));
+        clear();
+    }
+
+private:
+    std::vector<std::unique_ptr<CollectableBitBucket>> bin_buckets_;
+    std::vector<std::vector<char>> all_bin_bytes_;
```
klingaard, line 253: I'm having trouble understanding what this data structure is for.  It seems to not be really used or updated.

klingaard (commented):
More review to come... :)

sparta/sparta/collection/Collectable.hpp
```diff
@@ -30,26 +31,77 @@ namespace sparta{
...
-         */
+        template <typename T, typename = void>
+        struct use_raw_type : std::false_type {};
 
-        template<typename DataT, SchedulingPhase collection_phase = SchedulingPhase::Collection, typename = void>
-        class Collectable : public CollectableTreeNode
+        template <typename T>
+        struct use_raw_type<T,
+            std::enable_if_t<std::is_trivial_v<T> &&
+                             std::is_standard_layout_v<T>>> : std::true_type {};
```
klingaard: Consider moving to `requires`

sparta/sparta/collection/Collectable.hpp
```diff
@@ -128,50 +157,52 @@ namespace sparta{
...
-                std::ostringstream ss;
-                ss << val;
-                prev_annot_ = ss.str();
+            void initialize(const ValueType & val) {
+                //TODO cnyce: handle initial value
+                (void)val;
+            }
+
+            template <typename T>
+            std::enable_if_t<MetaStruct::is_any_pointer_v<T>, void>
```
klingaard: Consider using `requires`

sparta/sparta/collection/Collectable.hpp
```diff
@@ -337,558 +329,295 @@ namespace sparta{
...
-            //! The SchedulingPhase is ignored
-            void setManualCollection() {
-                auto_collect_ = false;
+                std::cout << "\nSerializing PairDefinition to database for '" << root_dtype << "'...\n";
+                for(size_t i = 0; i < names.size(); ++i) {
+                    std::cout << "\t" << names[i] << ", " << dtypes[i];
+                    if (!format_strings[i].empty()) {
+                        std::cout << " (" << format_strings[i] << ")";
+                    }
+                    std::cout << "\n";
```
klingaard: Can this be part of a verbosity mode?

sparta/sparta/pairs/SpartaKeyPairs.hpp
```diff
@@ -337,6 +392,13 @@ namespace sparta {
                 &pair_cache_, args...);
         }
 
+        /**
+         * \brief Use the given bit bucket to dump collected data to SimDB
+         */
+        void setBitBucket_(std::shared_ptr<collection::BitBucket> bit_bucket) {
```
klingaard: const &

sparta/sparta/pairs/SpartaKeyPairs.hpp
```diff
@@ -460,8 +522,16 @@ namespace sparta {
             return f_switch_;
         }
 
+        /**
+         * \brief Use the given bit bucket to dump collected data to SimDB
+         */
+        virtual void setBitBucket(std::shared_ptr<collection::BitBucket> bit_bucket) {
```
klingaard: const &

sparta/sparta/collection/BitBucket.hpp
colby-nyce: Yep - that's `EnumInspector` in SimDB:
https://github.com/sparcians/simdb/blob/main/include/simdb/apps/argos/EnumInspector.hpp

The enums are tracked in this call:
`enum_inspector_->inspect(val);`

And written to the DB at the end of simulation in `ArgosCollector::postTeardown()`

sparta/sparta/collection/BitBucket.hpp
colby-nyce: Yes, everything has to be a POD/enum/string at the end of the day. `SpartaKeyPairs.hpp` has the `BitBucket` and writes the values from the `SPARTA_ADDPAIR` getter functions one by one.

sparta/sparta/collection/BitBucket.hpp
colby-nyce: It's for `CollectableBitBucket`. I'll add `\class CollectableBitBucket` to make that more clear.

sparta/sparta/collection/BitBucket.hpp
colby-nyce: No they shouldn't since that bypasses things like string-to-int conversions. This has no business being public. Fixed.

sparta/sparta/collection/BitBucket.hpp
colby-nyce: Is the suggestion to add `all_bin_idxs_.reserve(capacity_);` to the constructor?

sparta/sparta/pairs/SpartaKeyPairs.hpp
```diff
@@ -1841,6 +1993,25 @@ namespace sparta {
         // Have a unique pointer to the ultimate terminal type of this KeyPairEntity.
         std::unique_ptr<ValueType> data_cpy_;
 
+        bool finalizeCollection_(
+            PairCache *& c, const ValueType & tmp) {
```
klingaard: can you use a better parameter name other than `c`?

sparta/sparta/pairs/SpartaKeyPairs.hpp
```diff
@@ -2262,10 +2468,20 @@ namespace sparta {
             return !was_clean;
         }
 
+        /**
+         * \brief Use the given bit bucket to dump collected data to SimDB
+         */
+        void setBitBucket(std::shared_ptr<collection::BitBucket> bit_bucket) {
```
klingaard: const &

sparta/sparta/utils/MetaStructs.hpp
klingaard, line 1: neat

sparta/sparta/collection/BitBucket.hpp
colby-nyce: Fixed

sparta/sparta/collection/BitBucket.hpp
colby-nyce: I changed all the vector::operator[] to use at(), and we'll see what it does to performance.

sparta/sparta/collection/BitBucket.hpp
colby-nyce: It holds the extracted bytes for each element in an `IterableCollector`. I made it a member variable so I wouldn't put a new one on the stack in the `writeTo()` method above.

sparta/sparta/collection/BitBucket.hpp
colby-nyce: Yes the template version is so we can do different things for strings/enums/PODs, and I've made the virtual `writeField` protected.

sparta/sparta/collection/BitBucket.hpp
colby-nyce: This may have been a leftover TODO -- I liberally used static_assert at first but was having trouble getting our simulators to compile. I bet this is no longer an issue, so I'll add it back and see if everything still compiles.

sparta/sparta/pairs/SpartaKeyPairs.hpp
colby-nyce: Changed to `cache`

sparta/sparta/pairs/SpartaKeyPairs.hpp
bdutro-mips, line 1177: I've been banging my head against some collection bugs and noticed an issue here, so since you're already under the hood...

This specialization doesn't actually work because the `std::is_same` clause is checking if the pointer type (`T`) is the same as the function argument type (`H`). The same bug is present in several places, which I'll note in other comments with "pointer bug here".

Once this is fixed, we then need to deal with a new problem, namely what do we want to do if we pass a null pointer to the collector?

bdutro-mips, line 1209: pointer bug here

sparta/sparta/pairs/SpartaKeyPairs.hpp
```diff
@@ -1556,18 +1711,7 @@ namespace sparta {
...
-                data_cpy_.reset(new ValueType(tmp));
-            }
-            else if(*data_cpy_ == tmp) {
-                return true;
-            }
-            updateValueInCache_(c, id_, tmp);
-            data_cpy_.reset(new ValueType(tmp));
-            return false;
+            return finalizeCollection_(c, tmp);
         }
```
bdutro-mips, line 1715: pointer bug here

sparta/sparta/pairs/SpartaKeyPairs.hpp
bdutro-mips, line 1747: pointer bug here

colby-nyce: Thanks for calling this out @bdutro-mips. How does the bug surface? Is it a wrong answer bug, or is it not collected at all, is it a crash, or does the wrong `populateFromEntityUtility_` get invoked? I would need a simple example using `PairDefinition` and `SPARTA_ADDPAIR` that draws the bug out. Even pseudo code will help. I just need to see an example of a collected data type that would hit this bug. If it's a hard crash, then a repro sim command would probably be sufficient.

Last question - is this an existing bug that you found using the v2 collector, or did you just find this while using my branch?

bdutro-mips: > How does the bug surface? Is it a wrong answer bug, or is it not collected at all, is it a crash, or does the wrong `populateFromEntityUtility_` get invoked?

One of the other `populateFromEntityUtility_` overloads is invoked (specifically one where the object type does not match the function/method type), which results in no collection if you pass a pointer to the collector without dereferencing it.

Something as simple as this should cover it:

```
class APairDef;

struct A
{
    using SpartaPairDefinitionType = APairDef;
    uint32_t getID() const { return 42; }
};

class APairDef
{
    APairDef() : PairDefinition<A>() { SPARTA_INVOKE_PAIRS(A); }
    SPARTA_REGISTER_PAIRS(SPARTA_ADDPAIR("id", &A::getID))
};

// later during simulation...

auto a = std::make_unique<A>(); // shared_ptr and SpartaSharedPointer should also trigger this bug
a_collector.collectWithDuration(a, 1);
```

I think the above version might only surface with Collectable objects, since the IterableCollector works a little differently. However, you can also trigger it with IterableCollectors if you `SPARTA_FLATTEN` a method that returns a pointer:

```
class BPairDef;

struct B
{
    using SpartaPairDefinitionType = BPairDef;
    std::unique_ptr a{new A};
    const auto& getA() const { return a; }
};

class BPairDef
{
    BPairDef() : PairDefinition<B>() { SPARTA_INVOKE_PAIRS(B); }
    SPARTA_REGISTER_PAIRS(SPARTA_FLATTEN(&B::getA))
};
```

> Last question - is this an existing bug that you found using the v2 collector, or did you just find this while using my branch?

It's an existing bug in the v2 collector

colby-nyce: @bdutro-mips I'm not seeing any weird behavior related to these bugs in the v3 collector (this PR branch). I did however see some head scratchers during development where a no-op `populateFromEntityUtility_` was being called, thus silently being unable to collect certain fields. I also saw a bug where calling `SPARTA_FLATTEN` on a method returning a pointer was unable to collect (I forget the bad code path so I don't know if it's the same bug you are seeing, but IIRC the field wasn't even flattened at all, just skipped over entirely, resulting in the front-end python code not being able to parse the bytes due to layout mismatches). Those were legitimate bugs with the v2 collector and were fixed in this PR. Not 100% sure if we are talking about the same bugs though.

Check out the new `populateFromEntityUtility_` methods that I added in this file, and check out the diff for `Collectable.hpp` to see if these changes seem like they would fix these bugs. If you have time, the best would be to run collection using this branch and see if the same issues persist. Otherwise I'm not sure what else I would be able to do at this point.

bdutro-mips: Ok, it sounds like you already ran into the issue then. Do you think it would be possible to remove the no-op `populateFromEntityUtility_` handlers and make them error at compile time or run time? That way we can avoid weirdness like this in the future.

colby-nyce: I'm not sure we should do that since all this SpartaKeyPairs code is also used for PEvents. Removing any of the no-op handlers causes the pair collection tests to stop compiling (ambiguous overloads). What I can do is differentiate between PEvents/Argos at runtime in the no-op handlers and throw if they are ever hit when using this code for Argos. Last thing I want to do is risk destabilizing map_v3 PEvents out of the gate.

bdutro-mips: Is there a use-case for no-op PEvent handlers? That sounds like another bug to me. However, I realize that's outside the original scope of this PR, so we could defer it for later.

klingaard (approved):

README.md
```diff
@@ -68,7 +63,7 @@ Use `apt`, `yum` or `brew` to install.
 
 Clone sparta:
 ```
-git clone --recursive git@github.com:sparcians/map --branch map_v2.2
+git clone --recursive git@github.com:sparcians/map --branch map_v3
```
knute-mips: Might want to have folks stick with v2 for now until simulators like Olympia are transitioned.  I'd like to consider map_v2.2 the most stable branch right now

README.md
```diff
@@ -27,22 +27,17 @@ MAP is broken into two parts:
 
 Current development branches:
 
-| map_v2.0                               | map_v2.1                                    | map_v2.2 |
-| -------------------------------------- | ------------------------------------------- | ---------|
-| Development prior to SimDB integration | SimDB integration/report generation support | TreeNode Extensions API Update |
+| map_v2.0                               | map_v2.1                                    | map_v2.2 | map_v3.0 |
+| -------------------------------------- | ------------------------------------------- | -------- | -------- |
+| Development prior to SimDB integration | SimDB integration/report generation support | TreeNode Extensions API Update | Redesigned Argos backed by SimDB |
```
knute-mips: It's more than a redesign -- it's a brand new Argos!!!

PR #684, opened 2026-09-14
Bump SimDB (Argos bug fixes / enhancements)

colby-nyce:
The SimDB changes have some fixes/enhancements that are needed to translate v2 layout files to the v3 format.

Scheduling lines widget:
- Support scalar structs / PODs / enums (not just queues)
- Force refresh widget on UI load (fixes wxpython "ghost" buttons that previously required a mouseover to make them disappear)
- Let users rename grid row labels (instead of "top.cpu.core0.fetch.foo" they could just call it "Foo Queue" etc., which is required to get the same layouts as before)

General:
- ALL_COLUMNS marker in layout yaml (easier to translate v2 layouts; without it, need to write out all collected struct fields in the new layouts manually which is rather tedious)
- Delete unused code


colby-nyce:
Don't bother reviewing yet - I found another issue with collected scalar enums. I'll fix it in SimDB and push to this branch soon.

colby-nyce:
All known bugs are fixed. Qualified by bashing all shipping layouts with DBs from real runs (regenerated for v3). Reopening PR.

klingaard (approved):

PR #685, opened 2026-09-15
Grow the PayloadEvent in-flight list on demand

ajainTT:
`PhasedPayloadEvent` grows its proxy pool on demand but tracks in-flight proxies in a `FastList` fixed at 1100, so the tracker can't follow the pool. The assert in `makeProxy_` only pre-empts `FastList`'s own out-of-room assert — a capacity artifact, not a modeled limit. A memory model under heavy backpressure reaches it legitimately: outstanding replay events are bounded by replay rate times delivery delay, not by any core structure.

Raising the constant is the obvious fix and the wrong one. The distribution is long-tailed — of a couple hundred PayloadEvents in one model, exactly one exceeds a handful outstanding — so a uniform ceiling sized for the tail charges every event the worst case.

### Change

`FastList` gains an `(initial, max)` constructor: reserve `max` once, construct `initial`, grow on demand. `PhasedPayloadEvent` and `PayloadEvent` take an optional `max_outstanding`.

The reserve is load bearing, not an optimisation: `Node` holds raw storage for `DataT`, so reallocation would byte-copy live elements. Reserving up front means `nodes_` never reallocates, so growth can't invalidate elements or iterators, and erase-during-iteration — which the cancel paths rely on — is unaffected.

Existing call sites are unchanged: `FastList(size)` delegates to `(size, size)` and `max_size()` still reports the ceiling, so `PriorityQueue` and the existing tests behave as before.

### Measured

| | before | after |
|---|---|---|
| RSS, 2000 lists | 52.1 MB | **9.6 MB** |
| Construct 2000 lists | 16.0 ms | **0.2 ms** |
| Hot path, 20M emplace+erase, equal capacity | 43.05 ms | 43.06 ms |

Only the nodes in use get constructed and touched, so RSS drops 5.4x. The hot path is unchanged — `ensureFreeNode_()` early-returns on a predicted branch where the assert sat. On-demand growth costs 0.6 ns per first-time node, paid once per node.

`reserve(max)` raises address space, not memory: VSZ goes 50.5 MB to 758 MB at 2000 lists, in untouched pages that never become resident and cost nothing under default overcommit. `max_outstanding` sizes it down where a VSZ limit applies.

### Tests

New `FastList` cases cover growth past the initial size, growth stopping at max, contents and iterators surviving a growth boundary, node reuse after erase, `initial == max` matching the old behaviour, bad arguments, and erase-during-iteration across a growth boundary.

Tested on Ubuntu with various gcc versions.

Please let me know what else I can do to get this over the finish line. Thank you.


ajainTT:
@klingaard and @colby-nyce, flagging this for review when you have a moment. It lifts the fixed 1100-entry ceiling on PhasedPayloadEvent's in-flight list by giving FastList an (initial, max) constructor that grows on demand, rather than raising the constant for every event. Thanks and please let me know if I can do anything to get this across the finish line.

sparta/sparta/utils/FastList.hpp
```diff
@@ -441,8 +460,23 @@ namespace sparta::utils
             return os;
         }
 
+        //! Ensure a free node is available, growing the list if below its ceiling.
+        void ensureFreeNode_() {
+            if(SPARTA_EXPECT_TRUE(free_head_ != -1)) { return; }
+            if(nodes_.size() < max_size_) {
```
klingaard: If `free_head_` is -1, then why do this check? I.e. wondering if it should just be an assert...

klingaard (approved):

klingaard:
Appreciate the submission!  This is a great extension to PLE.  Curious on the use case for it -- you mention this was related to memory accesses?

Also, the performance test -- I didn't see that in the tester.  Did you do that separately?

sparta/sparta/utils/FastList.hpp
ajainTT: Good catch, I have refactored the code with an assertion now.

ajainTT:
Yes, I tested it separately. Would you like me to add the performance tests to the repo too? I ran into PLE size assertions for a multi-core memory model.

sparta/sparta/utils/FastList.hpp
```diff
@@ -441,8 +460,21 @@ namespace sparta::utils
...
+            // Never reallocates: the ctor reserved max_size_.
+            const typename Node::NodeIdx new_idx =
+                static_cast<typename Node::NodeIdx>(nodes_.size());
+            nodes_.emplace_back(new_idx);
+            free_head_ = new_idx;
+        }
+
         // Stores all the nodes.
         std::vector<Node> nodes_;
+        size_t max_size_;     //!< Ceiling on the node count
```
colby-nyce: Probably should be const

colby-nyce (approved):
Great work and thanks for the very detailed PR notes!

sparta/sparta/utils/FastList.hpp
ajainTT: Thanks, I have made the fix. I think you'll have to re-run the workflows for this to be mergeable.

klingaard (approved):
Thank you for the submission.  This is a great addition.

klingaard:
I will merge these changes into branches map_v2.1 and map_v2.2

